PR #14, opened 2021-03-23
Trinket on-use/proc effects

Requielx:
Added most Pre-raid on-use and proc buff effects + Fix to Quagmirran's Eye (wasn't proccing/showing in buff section)


secretbis:
@Requielx This is excellent, thanks for the contribution!

PR #17, opened 2021-03-24
DST + Hourglass ICD

Requielx:
Added ICD to Dragonspine Trophy and Hourglass of the Unraveller based off datamined patch 2.5.1 (checked all other current proc trinket implemented and seem to match datamined info)


secretbis:
Hmm, they do indeed.  Great find, thank you!

PR #18, opened 2021-03-25
Added execute phase criterion

Requielx:
Couldn't see an existing rotation criteria for execute phase, discard if I missed it.


secretbis:
This is a great addition, thanks!

PR #23, opened 2021-03-27
RAP and Thundering Skyfire Diamond

Requielx:
- Updated items to include RAP that should of had RAP
- Updated Duration of Thundering Skyfire Diamond to 6 seconds (from 10) based on 2.5.1 data


secretbis:
Awesome, thanks!

PR #38, opened 2021-04-14
Heroic Strike Rank 11 is unavailable to players

Requielx:
Rank 10 is the max rank learnable by players, rank 11 was scrapped in TBC beta (original) and does not become available until wotlk


secretbis:
Great find, thanks!

PR #45, opened 2021-04-17
P1 Item Procs

Requielx:
Most of the proc rates are based on questionable wowhead comments or wowwiki, will need to update the proc rates when testing has been done. Blinkstrike still needs to be added to complete all p1 weapon procs


secretbis:
Excellent as usual, thank you!

PR #50, opened 2021-04-19
Multiple resource types for a participant, for Rogues/Druids

NoBetaBoredom:
I wanted to implement Rogues and saw that it's currently not possible to use multiple resources for one participant (Energy + CP for Rogues, or even Energy + CP + Mana for Druids).

So I added multiple resources, changed the relevant Criteria to include a resourceType (for RESOURCE_MISSING_GTE etc.) and a few other smaller things.

I never used Kotlin before and was unsure about a few things, let me know if this PR is acceptable or not.

### DIFF
--- a/src/commonMain/kotlin/character/Class.kt
+++ b/src/commonMain/kotlin/character/Class.kt
@@ -74,7 +74,7 @@ abstract class Class(
     abstract fun talentFromString(name: String, ranks: Int): Talent?
 
     // Class resource
-    abstract val resourceType: List<Resource.Type>
+    abstract val resourceTypes: List<Resource.Type>
 
     abstract val canDualWield: Boolean
 
--- a/src/commonMain/kotlin/character/Class.kt
+++ b/src/commonMain/kotlin/character/Class.kt
@@ -74,7 +74,7 @@ abstract class Class(
     abstract fun talentFromString(name: String, ranks: Int): Talent?
 
     // Class resource
-    abstract val resourceType: Resource.Type
+    abstract val resourceType: List<Resource.Type>
 
     abstract val canDualWield: Boolean
 
--- a/src/commonMain/kotlin/character/classes/boss/Boss.kt
+++ b/src/commonMain/kotlin/character/classes/boss/Boss.kt
@@ -14,7 +14,7 @@ class Boss(override var baseStats: Stats) : Class(mapOf(), BossSpec()) {
     }
 
     override var buffs: List<Buff> = listOf()
-    override var resourceType: Resource.Type = Resource.Type.MANA
+    override val resourceType: MutableList<Resource.Type> = mutableListOf(Resource.Type.MANA)
     override var canDualWield: Boolean = false
     override var attackPowerFromAgility: Int = 0
     override var attackPowerFromStrength: Int = 0
--- a/src/commonMain/kotlin/character/classes/boss/Boss.kt
+++ b/src/commonMain/kotlin/character/classes/boss/Boss.kt
@@ -14,7 +14,7 @@ class Boss(override var baseStats: Stats) : Class(mapOf(), BossSpec()) {
     }
 
     override var buffs: List<Buff> = listOf()
-    override val resourceType: MutableList<Resource.Type> = mutableListOf(Resource.Type.MANA)
+    override val resourceTypes: List<Resource.Type> = listOf(Resource.Type.MANA)
     override var canDualWield: Boolean = false
     override var attackPowerFromAgility: Int = 0
     override var attackPowerFromStrength: Int = 0
--- a/src/commonMain/kotlin/character/classes/rogue/specs/Combat.kt
+++ b/src/commonMain/kotlin/character/classes/rogue/specs/Combat.kt
@@ -1,0 +1,26 @@
+package character.classes.rogue.specs
+
+import character.Spec
+import character.SpecEpDelta
+
+class Combat : Spec() {
+    override val name: String = "Combat"
+    override val epBaseStat: SpecEpDelta = attackPowerBase
+    override val epStatDeltas: List<SpecEpDelta> = dualWieldMeleeDeltas
+    override val benefitsFromMeleeWeaponDps = true
+
+    override fun redSocketEp(deltas: Map<String, Double>): Double {
+        // 10 str
+        return (deltas["strength"] ?: 0.0) * 10.0
+    }
+
+    override fun yellowSocketEp(deltas: Map<String, Double>): Double {
+        // 10 crit rating
+        return (deltas["meleeCritRating"] ?: 0.0) * 10.0
+    }
+
+    override fun blueSocketEp(deltas: Map<String, Double>): Double {
+        // 5 str
+        return (deltas["strength"] ?: 0.0) * 5.0
+    }
+}
--- a/src/commonMain/kotlin/character/classes/rogue/specs/Combat.kt
+++ b/src/commonMain/kotlin/character/classes/rogue/specs/Combat.kt
@@ -10,17 +10,17 @@ class Combat : Spec() {
     override val benefitsFromMeleeWeaponDps = true
 
     override fun redSocketEp(deltas: Map<String, Double>): Double {
-        // 10 str
-        return (deltas["strength"] ?: 0.0) * 10.0
+        // 10 agi
+        return (deltas["agility"] ?: 0.0) * 10.0
     }
 
     override fun yellowSocketEp(deltas: Map<String, Double>): Double {
-        // 10 crit rating
-        return (deltas["meleeCritRating"] ?: 0.0) * 10.0
+        // 5 agi / 5 hit
+        return (deltas["meleeHitRating"] ?: 0.0) * 5 + (deltas["agility"] ?: 0.0) * 5
     }
 
     override fun blueSocketEp(deltas: Map<String, Double>): Double {
-        // 5 str
-        return (deltas["strength"] ?: 0.0) * 5.0
+        // 5 agi
+        return (deltas["agility"] ?: 0.0) * 5.0
     }
 }
--- a/src/commonMain/kotlin/character/classes/warlock/abilities/DemonicSacrifice.kt
+++ b/src/commonMain/kotlin/character/classes/warlock/abilities/DemonicSacrifice.kt
@@ -17,7 +17,7 @@ class DemonicSacrificeSuccubus : DemonicSacrifice("Succubus", { Stats(shadowDama
         const val name = "Demonic Sacrifice (Succubus)"
     }
 }
-class DemonicSacrificeFelguard : DemonicSacrifice("Felguard", { sp -> Stats(shadowDamageMultiplier = 1.15, manaPer5Seconds = (sp.getResource((Resource.Type.MANA)).maxAmount * 0.02 * 0.8).toInt())}) {
+class DemonicSacrificeFelguard : DemonicSacrifice("Felguard", { sp -> Stats(shadowDamageMultiplier = 1.15, manaPer5Seconds = ((sp.resources[Resource.Type.MANA]?.maxAmount ?: 0) * 0.02 * 0.8).toInt())}) {
     companion object {
         const val name = "Demonic Sacrifice (Felguard)"
     }
--- a/src/commonMain/kotlin/character/classes/warlock/abilities/DemonicSacrifice.kt
+++ b/src/commonMain/kotlin/character/classes/warlock/abilities/DemonicSacrifice.kt
@@ -3,6 +3,7 @@ package character.classes.warlock.abilities
 import character.Ability
 import character.Buff
 import character.Stats
+import character.Resource
 import sim.SimIteration
 import sim.SimParticipant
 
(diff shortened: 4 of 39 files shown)


src/commonMain/kotlin/character/classes/rogue/specs/Combat.kt
```diff
@@ -0,0 +1,26 @@
...
+
+class Combat : Spec() {
+    override val name: String = "Combat"
+    override val epBaseStat: SpecEpDelta = attackPowerBase
+    override val epStatDeltas: List<SpecEpDelta> = dualWieldMeleeDeltas
+    override val benefitsFromMeleeWeaponDps = true
+
+    override fun redSocketEp(deltas: Map<String, Double>): Double {
+        // 10 str
+        return (deltas["strength"] ?: 0.0) * 10.0
```
secretbis: 10 agility would be a rogue's best red gem, yes?

src/commonMain/kotlin/character/classes/rogue/specs/Combat.kt
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    override fun yellowSocketEp(deltas: Map<String, Double>): Double {
+        // 10 crit rating
+        return (deltas["meleeCritRating"] ?: 0.0) * 10.0
```
secretbis: This is pretty likely to be 5agi/5hit

src/commonMain/kotlin/character/classes/rogue/specs/Combat.kt
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    override fun blueSocketEp(deltas: Map<String, Double>): Double {
+        // 5 str
+        return (deltas["strength"] ?: 0.0) * 5.0
```
secretbis: This is likely to be 5 agi

src/commonMain/kotlin/sim/SimParticipant.kt
```diff
@@ -454,35 +458,56 @@ class SimParticipant(val character: Character, val rotation: Rotation, val sim:
...
+            amount = res.currentAmount.toDouble(),
+            delta = amount.toDouble(),
+            amountPct = res.currentAmount / res.maxAmount.toDouble() * 100.0,
+            resourceType = res.type,
+            abilityName = abilityName
+        ))
+    }
+
+    fun hasEnoughResource(type: Resource.Type, amount: Double) : Boolean {
+        if(amount == 0.0){return false} // for trinkets and other items that have MANA as the default
```
secretbis: Should this always return `true` if there is no resource cost?

src/commonMain/kotlin/character/classes/boss/Boss.kt
```diff
@@ -14,7 +14,7 @@ class Boss(override var baseStats: Stats) : Class(mapOf(), BossSpec()) {
     }
 
     override var buffs: List<Buff> = listOf()
-    override var resourceType: Resource.Type = Resource.Type.MANA
+    override val resourceType: MutableList<Resource.Type> = mutableListOf(Resource.Type.MANA)
```
secretbis: Do these need to be `MutableList`s?  I'd prefer if they could be non-mutable `List` and constructed with `listOf()`

src/commonMain/kotlin/character/classes/warlock/abilities/DemonicSacrifice.kt
```diff
@@ -16,7 +17,7 @@ class DemonicSacrificeSuccubus : DemonicSacrifice("Succubus", { Stats(shadowDama
         const val name = "Demonic Sacrifice (Succubus)"
     }
 }
-class DemonicSacrificeFelguard : DemonicSacrifice("Felguard", { sp -> Stats(shadowDamageMultiplier = 1.15, manaPer5Seconds = (sp.resource.maxAmount * 0.02 * 0.8).toInt())}) {
+class DemonicSacrificeFelguard : DemonicSacrifice("Felguard", { sp -> Stats(shadowDamageMultiplier = 1.15, manaPer5Seconds = (sp.getResource((Resource.Type.MANA)).maxAmount * 0.02 * 0.8).toInt())}) {
```
secretbis: Minor - there's a double paren around `Resource.Type.Mana`

src/commonMain/kotlin/sim/SimParticipant.kt
```diff
@@ -22,7 +22,7 @@ class SimParticipant(val character: Character, val rotation: Rotation, val sim:
     val logger = KotlinLogging.logger {}
 
     var stats: Stats = Stats()
-    lateinit var resource: Resource
+    lateinit var resource: MutableMap<Resource.Type, Resource>
```
secretbis: I think this type can be just `Map` - the object does need to be initially mutable to be created below, but the property can be the more generic non-mutable type.

src/commonMain/kotlin/sim/rotation/criteria/ResourceMissingGte.kt
```diff
@@ -15,7 +16,25 @@ class ResourceMissingGte(data: RotationRuleCriterion) : Criterion(Type.RESOURCE_
         null
     }
 
+    inline fun <reified T : Enum<T>> valueOf(type: String): T? {
+        return java.lang.Enum.valueOf(T::class.java, type)
```
secretbis: We cannot use Java types, as it would be non-portable to the JS build.  A companion method on `Resource` named `getResourceFromString(str: String): Resource?` or something like that would probably be best.

src/commonMain/kotlin/sim/rotation/criteria/ResourcePctGte.kt
```diff
@@ -15,7 +16,25 @@ class ResourcePctGte(data: RotationRuleCriterion) : Criterion(Type.RESOURCE_PCT_
         null
     }
 
+    inline fun <reified T : Enum<T>> valueOf(type: String): T? {
+        return java.lang.Enum.valueOf(T::class.java, type)
```
secretbis: We cannot use Java types, as it would be non-portable to the JS build.  A companion method on `Resource` named `getResourceFromString(str: String): Resource?` or something like that would probably be best.

src/commonMain/kotlin/sim/rotation/criteria/ResourcePctLte.kt
```diff
@@ -15,7 +16,25 @@ class ResourcePctLte(data: RotationRuleCriterion) : Criterion(Type.RESOURCE_PCT_
         null
     }
 
+    inline fun <reified T : Enum<T>> valueOf(type: String): T? {
+        return java.lang.Enum.valueOf(T::class.java, type)
```
secretbis: We cannot use Java types, as it would be non-portable to the JS build.  A companion method on `Resource` named `getResourceFromString(str: String): Resource?` or something like that would probably be best.

src/commonMain/kotlin/sim/SimParticipant.kt
```diff
@@ -90,8 +90,12 @@ class SimParticipant(val character: Character, val rotation: Rotation, val sim:
         // Compute initial stats
         recomputeStats()
 
-        // Initialize our subject resource
-        resource = Resource(this)
+        // Initialize our subject resource(s)
+        resource = mutableMapOf()
```
secretbis: We can use a local variable to build the map instead of assigning to the main property right away, and keep the main property non-mutable.

src/commonMain/kotlin/character/Class.kt
```diff
@@ -74,7 +74,7 @@ abstract class Class(
     abstract fun talentFromString(name: String, ranks: Int): Talent?
 
     // Class resource
-    abstract val resourceType: Resource.Type
+    abstract val resourceType: List<Resource.Type>
```
secretbis: Minor, as this is now a `List`, a better name might be `resourceTypes`.

src/commonMain/kotlin/sim/SimParticipant.kt
```diff
@@ -454,35 +458,56 @@ class SimParticipant(val character: Character, val rotation: Rotation, val sim:
...
+        if(amount == 0.0){return false} // for trinkets and other items that have MANA as the default
+
+        if(amount <= getResource(type).currentAmount) { 
+            return true 
+        }
+        return false
+    }
+
+    fun getResource(type: Resource.Type) : Resource {
+        return resource[type] ?: throw Exception("Tried to access a resource that is not present for participant")
```
secretbis: Hmm, I don't usually like to throw when (essentially) a map entry is not found - a caller could have a very reasonable way to recover from that.  

I think it would be better just having callers use the `resource` property directly, and removing this method.  I think there are enough ways that something might innocently try to get the current mana of a rogue or warrior or something, and this might end up too surprising.

It may reduce some brevity in callers, having to deal with `Resource?` instead of `Resource`, but I'd like to start there if we can.

secretbis (changes requested):
Thanks for the big contribution!  This is pretty close, I think, and definitely needed!

You've got a great start here, and let me know if you have any questions!

src/commonMain/kotlin/character/classes/rogue/specs/Combat.kt
NoBetaBoredom: Oh this wasn't even supposed to be in, I just copypasted the Warrior one before I realised I can't start on implementing Rogue before doing the resources thing. Changed to your recommendations for now.

src/commonMain/kotlin/character/classes/boss/Boss.kt
NoBetaBoredom: All changed to List now, all the characters too.

src/commonMain/kotlin/sim/SimParticipant.kt
NoBetaBoredom: Done.

src/commonMain/kotlin/sim/rotation/criteria/ResourceMissingGte.kt
NoBetaBoredom: `enumValueOf` is implemented in Kotlin for JS, changed it to that, should work. Do you still want it in an extra method in Resource?

src/commonMain/kotlin/sim/SimParticipant.kt
NoBetaBoredom: Solved.

src/commonMain/kotlin/character/Class.kt
NoBetaBoredom: Changed everywhere, while at it I also changed `resource` to `resources`.

src/commonMain/kotlin/sim/SimParticipant.kt
NoBetaBoredom: Yeah this was one of the things I was unsure about. I removed it and access `resources` directly now and handle null in the few places it was accessed. Still unsure how you want it to fail (silently? warn?), see examples below.

src/commonMain/kotlin/character/classes/warlock/abilities/DemonicSacrifice.kt
```diff
@@ -17,7 +17,7 @@ class DemonicSacrificeSuccubus : DemonicSacrifice("Succubus", { Stats(shadowDama
         const val name = "Demonic Sacrifice (Succubus)"
     }
 }
-class DemonicSacrificeFelguard : DemonicSacrifice("Felguard", { sp -> Stats(shadowDamageMultiplier = 1.15, manaPer5Seconds = (sp.getResource((Resource.Type.MANA)).maxAmount * 0.02 * 0.8).toInt())}) {
+class DemonicSacrificeFelguard : DemonicSacrifice("Felguard", { sp -> Stats(shadowDamageMultiplier = 1.15, manaPer5Seconds = ((sp.resources[Resource.Type.MANA]?.maxAmount ?: 0) * 0.02 * 0.8).toInt())}) {
```
NoBetaBoredom, line 20: This is an easy fail since it won't matter anyways if the participant has no Mana.

src/commonMain/kotlin/character/classes/warrior/abilities/Execute.kt
```diff
@@ -40,11 +40,13 @@ class Execute : Ability() {
 
     override fun cast(sp: SimParticipant) {
         val item = sp.character.gear.mainHand
-        val damage = 925.0 + (sp.getResource(resourceType(sp))).currentAmount * 21
+        val res = sp.resources[resourceType(sp)]
+        if(res == null) return
```
NoBetaBoredom: Here I don't know if it should just fail silently like this, it's effectively impossible anyways since `Rotation.next()` already makes sure the resource returned by `resourceType(sp)` exists. Still, if it was possible, how do you want that handled?

src/commonMain/kotlin/sim/SimParticipant.kt
NoBetaBoredom: Oops, yes

src/commonMain/kotlin/sim/rotation/criteria/ResourcePctLte.kt
```diff
@@ -34,7 +30,8 @@ class ResourcePctLte(data: RotationRuleCriterion) : Criterion(Type.RESOURCE_PCT_
     }
 
     override fun satisfied(sp: SimParticipant): Boolean {
-        if(pct == null || resourceType == null) {return false}
-        return sp.getResource(resourceType).currentAmount <= sp.getResource(resourceType).maxAmount * (pct / 100.0)
+        val res = sp.resources[resourceType]
+        if(pct == null || resourceType == null || res == null) {return false}
```
NoBetaBoredom: Same here, print a warning or no?

src/commonMain/kotlin/character/classes/warrior/abilities/Execute.kt
secretbis: In cases like this where it should never fail (and deserves a crash when it does), we can just force the compiler to consider it non-nullable with `!!`

src/commonMain/kotlin/sim/rotation/criteria/ResourcePctLte.kt
secretbis: Given that criteria consume user input, we should probably warn and return false - the most likely cause is a typo in a rotation.

secretbis (approved):
Great work!  Thanks for the changes!

PR #51, opened 2021-04-22
P1 and Legacy Trinkets

Requielx:
-Added the rest of the P1 Trinkets (DPS only)
-Added some higher end Legacy trinkets that may be used in TBC (Except the black book due to warlock pets not implemented yet)
-Removed double entry for polearm Normalisation (oops)
-Tidied up buffs.kt so its split between weapons/trinket procs and is now is order of spell ID


src/commonMain/kotlin/data/buffs/PendantOfTheVioletEye.kt
```diff
@@ -0,0 +1,72 @@
...
+        override val name: String = "Pendant of the Violet Eye"
+        override fun gcdMs(sp: SimParticipant): Int = 0
+        override fun cooldownMs(sp: SimParticipant): Int = 120000
+
+        override fun trinketLockoutMs(sp: SimParticipant): Int = buffDurationMs
+
+        override fun cast(sp: SimParticipant) {
+            val state = stackBuff.state(sp)
+            sp.addBuff(buff)
+            if(state.currentStacks > 1){
```
secretbis: We shouldn't need to reset this here - stacks will automatically reset when the buff is reset on expiration.  

https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/sim/SimParticipant.kt#L332
https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/character/Buff.kt#L73

src/commonMain/kotlin/data/buffs/RomulosPoisonVial.kt
```diff
@@ -0,0 +1,67 @@
...
+            Trigger.RANGED_YELLOW_HIT,
+            Trigger.RANGED_YELLOW_CRIT,
+            Trigger.RANGED_BLOCK,
+        )
+        override val type: Type = Type.PPM
+        override val ppm: Double = 2.0
+        //TODO: test proc rate on live, value based on wowhead comments with a sample size of 1k
+
+        override fun proc(sp: SimParticipant, items: List<Item>?, ability: Ability?, event: Event?) {
+                natureAbility.cast(sp)
```
secretbis: Minor - these three lines are double-indented

src/commonMain/kotlin/data/buffs/PendantOfTheVioletEye.kt
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        override fun modifyStats(sp: SimParticipant): Stats? {
+            val state = stackBuff.state(sp)
+            return Stats(manaPer5Seconds = 21 * state.currentStacks)
+        }
+
+        override fun procs(sp: SimParticipant): List<Proc> = listOf(proc)
+    }
+    val stackBuff = object : Buff() {
```
secretbis: We can probably merge this with the inline buff declaration above - the sim should handle the state management.

src/commonMain/kotlin/data/buffs/BadgeOfTheSwarmguard.kt
```diff
@@ -0,0 +1,88 @@
...
+                Trigger.RANGED_YELLOW_CRIT,
+                Trigger.RANGED_BLOCK,
+            )
+            override val type: Type = Type.PPM
+            override val ppm: Double = 10.0
+            //from fightclub discord
+            override fun cooldownMs(sp: SimParticipant): Int = 0
+            override fun proc(sp: SimParticipant, items: List<Item>?, ability: Ability?, event: Event?){
+                val state = stackBuff.state(sp)
+                if (state.currentStacks < stackBuff.maxStacks) {
```
secretbis: We can remove this check - the Buff internals should handle max stacks:

https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/character/Buff.kt#L60

src/commonMain/kotlin/data/buffs/BadgeOfTheSwarmguard.kt
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        override fun modifyStats(sp: SimParticipant): Stats? {
+            val state = stackBuff.state(sp)
+            return Stats(armorPen= 200 * state.currentStacks)
+        }
+
+        override fun procs(sp: SimParticipant): List<Proc> = listOf(proc)
+    }
+    val stackBuff = object : Buff() {
```
secretbis: We can probably combine this with the inline buff on L21, and set `maxStacks` in there.  Helps with stats if the average stacks are in the state of a non-hidden buff.

secretbis (changes requested):
Looks great!  Just had a couple comments - let me know what you think!

Requielx:
Yeah don't know what I was thinking with having the stack component outside the on-use buff. Thanks for picking that up @marisa-ashkandi

src/commonMain/kotlin/data/buffs/BadgeOfTheSwarmguard.kt
```diff
@@ -0,0 +1,81 @@
...
+    override val name: String = Companion.name
+    override val durationMs: Int = -1
+    override val hidden: Boolean = true
+
+    val buffDurationMs = 30000
+    val buff = object : Buff() {
+        override val name: String  = "Badge of the Swarmguard"
+        override val durationMs: Int = buffDurationMs
+
+        val stackBuff = object : Buff() {
```
secretbis, line 25: We can still combine this inline buff with the containing one.  Currently, these two are redundant, and we should be able to remove `stackBuff`.  

We can rename L22 to "Insight of the Qiraji", add the `maxStacks` config to that Buff object.  Then, the state check on L60 can reference `this`'s state object, instead of `stackBuff`.  

The Kotlin type system will get a little confused, since the proc on L32 would be adding `this` Buff to the sim, but `this` technically doesn't exist yet since the Proc construction is contained within that same Buff object.  

Here is an example of how this can be worked around - just pass `this` to a function outside the `Buff` declaration that returns a `Proc`.  https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/character/classes/shaman/abilities/ElementalMastery.kt#L49

src/commonMain/kotlin/data/buffs/PendantOfTheVioletEye.kt
```diff
@@ -0,0 +1,64 @@
...
+    override val name: String = Companion.name
+    override val durationMs: Int = -1
+    override val hidden: Boolean = true
+
+    val buffDurationMs = 20000
+    val buff = object : Buff() {
+        override val name: String  = "Pendant of the Violet Eye"
+        override val durationMs: Int = buffDurationMs
+
+        val stackBuff = object : Buff() {
```
secretbis, line 25: This can be rewritten similarly to Badge, where we can use the L22 buff alone.

src/commonMain/kotlin/data/buffs/BadgeOfTheSwarmguard.kt
```diff
@@ -0,0 +1,81 @@
...
+            override val name: String = "Insight of the Qiraji"
+            override val durationMs: Int = -1
```
secretbis, line 27: I'm not sure this would ever be removed, so stacks on repeated Badge activations might remember the previous stack state (and be fully-stacked instantly).  With the suggested refactor above, we shouldn't have to worry about that since the single Buff will have a duration, and be auto-removed on expiration.

secretbis (changes requested):
Thanks for the changes!  I think we can simplify it a little further - I added a link to an example of how to nest a Proc within a Buff when the Proc needs to reference that same Buff object.  This should eliminate the need for the two extra `stackBuff` Buffs in Badge and Violet Eye.

secretbis (approved):

PR #53, opened 2021-05-01
Lionheart PPM and Blinkstrike added

Requielx:
- Updated lionheart PPM based on beta testing 3.0 -> 1.0
- Added blinkstrikes proc


secretbis:
Looks great, thanks!

PR #54, opened 2021-05-03
Rogue first pass

NoBetaBoredom:
I might have went a bit overboard and added **all** the Rogue things (all simrelevant abilities/talents).
Most abilities/talents are still untested and some damage calculations are probably wrong. I did test most of the things in the combat preset below. Also found a few bugs that affect all melee.

I will continue to test and refine things over the days.

I went ahead and made a basic [rogue_combat_preraid.yml](https://gist.github.com/NoBetaBoredom/0b4bb597fec1c0428e9883e78d1a38fc) preset. The rotation is pretty basic right now and can definitely be improved, I'm still thinking about the best way to do a bit more advanced things.


secretbis:
@NoBetaBoredom Wow!  I'll be able to test this in the next day or so, thanks for this!

src/commonMain/kotlin/character/classes/rogue/talents/DirtyDeeds.kt
```diff
@@ -0,0 +1,84 @@
...
+            )
+        }
+    }
+
+    val buff = object : Buff() {
+        override val name: String = "${Companion.name} (Talent)"
+        override val durationMs: Int = -1
+        override val hidden: Boolean = true
+
+        var appliedExecuteBonusBuff = false
```
secretbis: State should only be stored on the SimParticipant state collections, by using the `state()` accessor on `Buff`.  Talent objects are only constructed once per sim run, so this state variable here wouldn't be thread-safe.  

Here's an example of a similar operation: 

Windfury Weapon ICD check: https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/character/classes/shaman/buffs/WindfuryWeapon.kt#L56
WF Weapon custom buff state object: https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/character/classes/shaman/buffs/WindfuryWeapon.kt#L12

I see the issue here with buffs that behave differently depending on target health - it wouldn't get a recompute at the right time.  I'll think on how to improve that.

src/commonMain/kotlin/character/classes/rogue/talents/FindWeakness.kt
```diff
@@ -0,0 +1,53 @@
...
+    fun damageMultiplier(): Double {
+        return 1.0 + (currentRank * 0.02)
+    }
+
+    val dmgBuff = object : Buff() {
+        override val name: String = "Find Weakness"
+        override val durationMs: Int = 10000
+
+        override fun modifyStats(sp: SimParticipant): Stats {  
+            // unsure if yellowDamageMultiplier or yellowDamageFlatModifier
```
secretbis, line 26: Yes, `yellowDamageMultiplier` is correct here - it's for percentage modifiers.  `yellowDamageFlatModifier` is any non-percentage, static damage bonus to yellow damage

src/commonMain/kotlin/character/classes/rogue/talents/Vigor.kt
```diff
@@ -0,0 +1,38 @@
...
+    fun maxEnergyIncrease(): Int {
+        return currentRank * 10
+    }
+
+    val buff = object : Buff() {
+        override val name: String = "${Companion.name} (Talent)"
+        override val durationMs: Int = -1
+        override val hidden: Boolean = true
+
+        var energyGiven = false
```
secretbis: This bit of in-buff state has similar thread-safety issues as the prior comment.  

That said, there is also a sort of paradoxical `SimParticpant` construction issue here.  Resources aren't initialized until the initial set of buffs from gear/talents/precombat/etc are cast, mostly for computing mana.  So, `sp.resources` is uninitialized at this point in time.  

It would be pretty hard to handle Vigor as a Buff, as currently implemented.  The simplest way is probably just to check the talent presence in `Resource.max()`, as it gets a `SimParticpant` reference to use.

src/commonMain/kotlin/data/gems/ChaoticSkyfireDiamond.kt
```diff
@@ -16,6 +16,7 @@ class ChaoticSkyfireDiamond : Gem(34220, "Chaotic Skyfire Diamond", "inv_misc_ge
         override val durationMs: Int = -1
         override val hidden: Boolean = true
 
+        // TODO: on my quest for critdmg calculation i found out this critMultiplier applies to the full dmg of the crit, not just the bonus damage. this has to be applied to the base critMultiplier (Stats.physicalCritMultiplier etc.) before handling multipliers
```
secretbis, line 19: Thanks for pointing that out!  Filed #56

src/commonMain/kotlin/sim/SimParticipant.kt
```diff
@@ -638,12 +647,13 @@ class SimParticipant(val character: Character, val rotation: Rotation, val sim:
         return stats.expertiseRating / Rating.expertisePerPct
     }
 
+    // from the numbers that are displayed ingame, you also have to subtract the race-specific agility bonuses from this to get an accurate result
```
secretbis, line 650: I'm not sure what you mean here - can you elaborate on the mechanics?  I also see a strength reduction by 10 on L619 above - is that intentional?

src/commonMain/kotlin/mechanics/Spell.kt
```diff
@@ -125,7 +128,8 @@ object Spell {
     // Performs an attack roll given an initial unmitigated damage value
     fun attackRoll(sp: SimParticipant, damageRoll: Double, school: Constants.DamageType, isBinary: Boolean = false, bonusCritChance: Double = 0.0, bonusHitChance: Double = 0.0) : Pair<Double, Event.Result> {
         // Find all our possible damage mods from buffs and so on
-        val critMultiplier = Stats.spellCritMultiplier + (sp.stats.spellDamageAddlCritMultiplier - 1)
+        // this formula was wrong, bonus should only apply to the critbonus part of the damage, this is the general version
```
secretbis, line 131: Thanks for the fix!

src/commonMain/kotlin/data/items/ItemIndex.kt
```diff
@@ -3873,7 +3873,8 @@ public object ItemIndex {
       { BrightbrewCharm() },
       { DirebrewsShanker() },
       { BalebrewCharm() },
-      { TheHorsemansBlade() }
+      { TheHorsemansBlade() },
+      { NightstalkersWristguards() }
```
secretbis: This file gets regenerated - any one-off items should be implemented in `Items.kt`

src/commonMain/kotlin/data/items/NightstalkersWristguards.kt
```diff
@@ -0,0 +1,61 @@
+package `data`.items
```
secretbis: Would you mind moving this item into the `itemscustom` package (next to [Annihilator](https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/data/itemscustom/Annihilator.kt) and others).  This package is intended to be nuked and regenerated by `ItemGen.kt`

src/commonMain/kotlin/mechanics/Melee.kt
```diff
@@ -125,31 +127,33 @@ object Melee {
...
+    fun meleeDodgeChance(sp: SimParticipant, item: Item?): Double {
         return (valueByLevelDiff(sp, baseDodgeChance) - (sp.expertisePct() / 100.0) - (expertisePctForItem(sp, item) / 100.0)).coerceAtLeast(0.0)
     }
 
     fun meleeGlanceChance(sp: SimParticipant): Double {
         return valueByLevelDiff(sp, baseGlancingChance)
     }
 
-    fun meleeGlanceReduction(sp: SimParticipant, item: Item): Double {
+    // this is actually the meleeGlanceMultiplier and not the reduction.
```
secretbis, line 152: Yikes, that's impactful.  Thanks for finding that!

src/commonMain/kotlin/mechanics/Melee.kt
```diff
@@ -178,14 +182,40 @@ object Melee {
...
+
+    fun additionalWeaponTypeCritChance(sp: SimParticipant, item: Item?): Double {
+        return when(item?.itemSubclass) {
+            Constants.ItemSubclass.DAGGER -> sp.stats.daggerAdditionalCritChancePercent
+            Constants.ItemSubclass.FIST -> sp.stats.fistWeaponAdditionalCritChancePercent
+            else -> 0.0
+        }
+    }
+
+    /*  maybe do this instead of the optional arguments I added below for crit, critdmg and dodge?
```
secretbis, line 200: I would probably shy away from doing that at the moment, considering the mutability of the object, and the expense of recalculating.  

At one point in time, the `Stats` object did have a mirror `subtract()` method, but I never had a use for it, and eventually removed it.  Perhaps it could be added back to make this kind of operation efficient enough.

src/commonMain/kotlin/character/auto/MeleeMainHandRogue.kt
```diff
@@ -0,0 +1,22 @@
+package character.auto
+
+import data.model.Item
+import sim.SimParticipant
+import character.classes.rogue.buffs.*
+
+class MeleeMainHandRogue : AutoAttackBase() {
```
secretbis: Could this extend `MeleeMainHand` instead, and delegate the `sp.hasMainHandWeapon()` check below?

src/commonMain/kotlin/character/auto/MeleeOffHandRogue.kt
```diff
@@ -0,0 +1,22 @@
+package character.auto
+
+import data.model.Item
+import sim.SimParticipant
+import character.classes.rogue.buffs.*
+
+class MeleeOffHandRogue : AutoAttackBase() {
```
secretbis: Could this extend `MeleeOffHand` instead, and delegate the `sp. isDualWielding()` check below?

src/commonMain/kotlin/character/classes/rogue/Rogue.kt
```diff
@@ -2,23 +2,124 @@ package character.classes.rogue
...
-        intellect = 123,
-        strength = 209,
-        stamina = 154,
-        spirit = 92
+        agility = 158,
+        intellect = 39,
+        strength = 95,
+        stamina = 90,
+        spirit = 58,
+        attackPower = 140
```
secretbis, line 16: Rogues have base `attackPower`?  Interesting.

src/commonMain/kotlin/character/classes/rogue/abilities/AdrenalineRush.kt
```diff
@@ -0,0 +1,41 @@
...
+    override val castableOnGcd: Boolean = true
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 0.0
+
+    override fun available(sp: SimParticipant): Boolean {
+        val ar = sp.character.klass.talents[character.classes.rogue.talents.AdrenalineRush.name] as character.classes.rogue.talents.AdrenalineRush?
+        val available = if(ar != null){ ar.currentRank == ar.maxRank } else { false }
+        if (!available) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having the corresponding talent" }
```
secretbis: I'd prefer to downgrade the log level here to `debug`, as this would be very verbose.  Omitting entirely is also fine.

src/commonMain/kotlin/character/classes/rogue/abilities/Ambush.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    override fun gcdMs(sp: SimParticipant): Int = sp.physicalGcd().toInt()
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 60.0
+
+    override fun available(sp: SimParticipant): Boolean {
+        val usesDagger = sp.character.gear.mainHand.itemSubclass == Constants.ItemSubclass.DAGGER
+        if (!usesDagger) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having a dagger in the mainhand" }
```
secretbis: I'd prefer to downgrade the log level here to `debug`, as this would be very verbose.  Omitting entirely is also fine.

src/commonMain/kotlin/character/classes/rogue/abilities/Ambush.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    val bonusDamage = 335
+    val weaponDamageMultiplier = 2.75
+    override fun cast(sp: SimParticipant) {
+
+        var increasedDamagePercent = 0.0
+        val opportunity = sp.character.klass.talents[Opportunity.name] as Opportunity?
+        increasedDamagePercent += opportunity?.damageIncreasePercent() ?: 0.0
+        
+        val dmgMultiplier = 1 + (increasedDamagePercent / 100.0).coerceAtLeast(0.0)
```
secretbis, line 43: Would it be better to have Opportunity return a multiplier instead?  It seems that would simplify code in a number of abilities.  

The bit about converting it to a multiplier seems to be duplicated across the abilities it affects.

src/commonMain/kotlin/character/classes/rogue/abilities/Backstab.kt
```diff
@@ -0,0 +1,86 @@
...
+    val weaponDamageMultiplier = 1.50
+    override fun cast(sp: SimParticipant) {
+
+        val pw = sp.character.klass.talents[PuncturingWounds.name] as PuncturingWounds?
+        val increasedCritChance = pw?.additionalCritChanceBackstab() ?: 0.0
+
+        val lethality = sp.character.klass.talents[Lethality.name] as Lethality?
+        val critDmgMultiplier = lethality?.critDamageMultiplier() ?: 1.0
+
+        val aggression = sp.character.klass.talents[Aggression.name] as ImprovedEviscerate?
```
secretbis: This is being typecast to the wrong type

src/commonMain/kotlin/character/classes/rogue/abilities/Backstab.kt
```diff
@@ -0,0 +1,86 @@
...
+        var increasedDamagePercent = aggression?.damageIncreasePercent() ?: 0.0
+        val surprise = sp.character.klass.talents[SurpriseAttacks.name] as SurpriseAttacks?
+        increasedDamagePercent += surprise?.damageIncreasePercent() ?: 0.0
```
secretbis, line 46: Do all these rogue talents stack additively, instead of multiplicatively?

src/commonMain/kotlin/character/classes/rogue/abilities/BladeFlurry.kt
```diff
@@ -0,0 +1,42 @@
...
+    override fun gcdMs(sp: SimParticipant): Int = sp.physicalGcd().toInt()
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 25.0
+
+    override fun available(sp: SimParticipant): Boolean {
+        val bf = sp.character.klass.talents[character.classes.rogue.talents.BladeFlurry.name] as character.classes.rogue.talents.BladeFlurry?
+        val available = if(bf != null){ bf.currentRank == bf.maxRank } else { false }
+        if (!available) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having the corresponding talent" }
```
secretbis: I'd prefer to downgrade the log level here to debug, as this would be very verbose. Omitting entirely is also fine.

src/commonMain/kotlin/character/classes/rogue/abilities/Backstab.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    override fun gcdMs(sp: SimParticipant): Int = sp.physicalGcd().toInt()
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 60.0
+
+    override fun available(sp: SimParticipant): Boolean {
+        val usesDagger = sp.character.gear.mainHand.itemSubclass == Constants.ItemSubclass.DAGGER
+        if (!usesDagger) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having a dagger in the mainhand" }
```
secretbis: I'd prefer to downgrade the log level here to debug, as this would be very verbose. Omitting entirely is also fine.

src/commonMain/kotlin/character/classes/rogue/abilities/ColdBlood.kt
```diff
@@ -0,0 +1,43 @@
...
+    override val castableOnGcd: Boolean = true
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 0.0
+
+    override fun available(sp: SimParticipant): Boolean {
+        val cb = sp.character.klass.talents[character.classes.rogue.talents.ColdBlood.name] as character.classes.rogue.talents.ColdBlood?
+        val available = if(cb != null){ cb.currentRank == cb.maxRank } else { false }
+        if (!available) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having the corresponding talent" }
```
secretbis: I'd prefer to downgrade the log level here to debug, as this would be very verbose. Omitting entirely is also fine.

src/commonMain/kotlin/character/classes/rogue/abilities/DeadlyPoison.kt
```diff
@@ -0,0 +1,29 @@
...
+    override val id: Int = 27187
+    override fun gcdMs(sp: SimParticipant): Int = 0
+    override val castableOnGcd: Boolean = true
+
+    override fun available(sp: SimParticipant): Boolean {
+        return if(Melee.isOffhand(sp, item)) { sp.isDualWielding() } else true
+    }
+
+    override fun cast(sp: SimParticipant) {
+        // TODO: this needs to be casted so it can be resisted.
```
secretbis: So this needs a single resistance roll, but it doesn't act like a "spell" in terms of opposed hit checks?  I think that should be possible to factor that out of the spell code.

I'm happy to take that on after this is merged!  In the absence of boss resistance data and ways to select it in the sim, I think this should be fine to leave as-is for now.

src/commonMain/kotlin/character/classes/rogue/abilities/Envenom.kt
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        val vp = sp.character.klass.talents[VilePoisons.name] as VilePoisons?
+        val dmgIncrease = vp?.damageIncreasePercentEnvenom() ?: 0.0
+
+        val dmgMultiplier = 1 + (dmgIncrease / 100.0).coerceAtLeast(0.0)
+
+        // TODO: not sure if this really is just a flat increase on the final damage
+        val damageRoll = damage * dmgMultiplier
+        val result = Melee.attackRoll(sp, damageRoll, null, isWhiteDmg = false, noDodgeAllowed = noDodgeAllowed(sp))
```
secretbis, line 66: Is Envenom affected by nature resistance?  I am wondering if this needs a resist roll similar to what Deadly Poison needs.

src/commonMain/kotlin/character/classes/rogue/abilities/Envenom.kt
```diff
@@ -0,0 +1,106 @@
...
+            else -> null
+        }
+
+        fireProcAsFinisher(sp, triggerTypes, null, event)
+    }
+
+    fun maxPoisonDosesAvailable(sp: SimParticipant): Int {
+        val debuffState = sp.sim.target.debuffState[DeadlyPoisonDot.name]
+        if(debuffState == null) {
+            KotlinLogging.logger{}.warn{ "Tried to cast $name but there was no deadly poison debuff present" }
```
secretbis: I'd prefer to downgrade the log level here to debug, as this would be very verbose. Omitting entirely is also fine.

Same for the below message on L96

src/commonMain/kotlin/character/classes/rogue/abilities/Envenom.kt
```diff
@@ -0,0 +1,106 @@
...
+            return 0
+        }
+        return debuffState.currentStacks+1
```
secretbis: Would this be `currentStacks` instead of stacks plus one?

src/commonMain/kotlin/character/classes/rogue/abilities/Garrote.kt
```diff
@@ -0,0 +1,69 @@
...
+    override fun available(sp: SimParticipant): Boolean {
+        val inStealth = sp.buffs[Stealth.name] != null
+        return inStealth && super.available(sp)
+    }
+
+    override fun cast(sp: SimParticipant) {
+        // can't be dodged, parried or blocked. parry/block implied by sp.sim.opts.allowParryAndBlock
+        val result = Melee.attackRoll(sp, 0.0, null, isWhiteDmg = false, noDodgeAllowed = true)
+
+        val event = Event(
```
secretbis, line 41: We can omit the damage event for Garrote - applying the dot and firing on-hit procs should be all we need to do.

src/commonMain/kotlin/character/classes/rogue/abilities/GhostlyStrike.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun gcdMs(sp: SimParticipant): Int = sp.physicalGcd().toInt()
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 40.0
+
+    override fun available(sp: SimParticipant): Boolean {
+        val gs = sp.character.klass.talents[character.classes.rogue.talents.GhostlyStrike.name] as character.classes.rogue.talents.GhostlyStrike?
+        val available = if(gs != null){ gs.currentRank == gs.maxRank } else { false }
+        if (!available) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having the corresponding talent" }
```
secretbis: Similar to other reduce-log-level-or-omit comments.

src/commonMain/kotlin/character/classes/rogue/abilities/Hemorrhage.kt
```diff
@@ -0,0 +1,78 @@
...
+    override fun gcdMs(sp: SimParticipant): Int = sp.physicalGcd().toInt()
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 40.0
+
+    override fun available(sp: SimParticipant): Boolean {
+        val hemo = sp.character.klass.talents[character.classes.rogue.talents.Hemorrhage.name] as character.classes.rogue.talents.Hemorrhage?
+        val available = if(hemo != null){ hemo.currentRank == hemo.maxRank } else { false }
+        if (!available) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having the corresponding talent" }
```
secretbis: Similar to other reduce-log-level-or-omit comments.

src/commonMain/kotlin/character/classes/rogue/talents/VilePoisons.kt
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    override val name: String = Companion.name
+    override val maxRank: Int = 5
+
+    fun damageIncreasePercentEnvenom(): Double {
+        return currentRank * 4.0
+    }
+
+    fun damageIncreasePercentPoisons(): Double {
```
secretbis, line 17: Possible to combine this and the Envenom method?

src/commonMain/kotlin/character/classes/rogue/abilities/Hemorrhage.kt
```diff
@@ -0,0 +1,78 @@
...
+        val event = Event(
+            eventType = Event.Type.DAMAGE,
+            damageType = Constants.DamageType.PHYSICAL,
+            abilityName = name,
+            amount = result.first,
+            result = result.second,
+        )
+        sp.logEvent(event)
+
+        if(result.second != Event.Result.MISS && result.second != Event.Result.DODGE) {
```
secretbis: This should also account for parry

src/commonMain/kotlin/character/classes/rogue/abilities/GhostlyStrike.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun cast(sp: SimParticipant) {
+
+        val lethality = sp.character.klass.talents[Lethality.name] as Lethality?
+        val critDmgMultiplier = lethality?.critDamageMultiplier() ?: 1.0
+
+        val item = sp.character.gear.mainHand
+        val damageRoll = Melee.baseDamageRoll(sp, item) * weaponDamageMultiplier
+        val result = Melee.attackRoll(sp, damageRoll, item, isWhiteDmg = false, abilityAdditionalCritDamageMultiplier = critDmgMultiplier)
+        
+        if(result.second != Event.Result.MISS && result.second != Event.Result.DODGE) {
```
secretbis: This should account for parry

src/commonMain/kotlin/character/classes/rogue/abilities/Mutilate.kt
```diff
@@ -0,0 +1,138 @@
...
+    override fun gcdMs(sp: SimParticipant): Int = sp.physicalGcd().toInt()
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 60.0
+
+    override fun available(sp: SimParticipant): Boolean {
+        val mutilate = sp.character.klass.talents[character.classes.rogue.talents.Mutilate.name] as character.classes.rogue.talents.Mutilate?
+        val available = if(mutilate != null){ mutilate.currentRank == mutilate.maxRank } else { false }
+        if (!available) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having the corresponding talent" }
```
secretbis: Similar to other reduce-log-level-or-omit comments.

src/commonMain/kotlin/character/classes/rogue/abilities/Mutilate.kt
```diff
@@ -0,0 +1,138 @@
...
+            else -> null
+        }
+
+        if(triggerTypes != null) {
+            sp.fireProc(triggerTypes, listOf(item), this, event)
+        }
+    }
+
+    fun castOffhand(sp: SimParticipant, bonusDamage: Double) {
+        val opportunity = sp.character.klass.talents[Opportunity.name] as Opportunity?
```
secretbis, line 105: Puncturing Wounds and Lethality should also apply to the offhand hit

src/commonMain/kotlin/character/classes/rogue/talents/DirtyDeeds.kt
```diff
@@ -0,0 +1,84 @@
+package character.classes.rogue.talents
+
+import character.*
+import sim.SimParticipant
+import data.model.Item
+import sim.Event
+
+// TODO: need to find a better way for this
```
secretbis, line 8: This could be just handled in each ability - a little tedious, but would work fine.

src/commonMain/kotlin/character/classes/rogue/abilities/Hemorrhage.kt
```diff
@@ -0,0 +1,78 @@
...
+    override fun cast(sp: SimParticipant) {
+
+        val lethality = sp.character.klass.talents[Lethality.name] as Lethality?
+        val critDmgMultiplier = lethality?.critDamageMultiplier() ?: 1.0
+
+        val item = sp.character.gear.mainHand
+        val damageRoll = Melee.baseDamageRoll(sp, item, isNormalized = true) * weaponDamageMultiplier
+        val result = Melee.attackRoll(sp, damageRoll, item, isWhiteDmg = false, abilityAdditionalCritDamageMultiplier = critDmgMultiplier)
+        
+        if(result.second != Event.Result.MISS && result.second != Event.Result.DODGE) {
```
secretbis, line 45: This should account for parry

src/commonMain/kotlin/character/classes/rogue/abilities/Mutilate.kt
```diff
@@ -0,0 +1,138 @@
...
+        if (!usesDaggers) {
+            KotlinLogging.logger{}.warn{ "Tried to use ability $name without having a daggers in both hands" }
+        }
+
+        return available && usesDaggers && super.available(sp)
+    }
+
+    fun castMainhand(sp: SimParticipant, bonusDamage: Double) {
+
+        val pw = sp.character.klass.talents[PuncturingWounds.name] as PuncturingWounds?
```
secretbis, line 49: Opportunity should also apply to the mainhand hit

src/commonMain/kotlin/character/classes/rogue/abilities/Mutilate.kt
```diff
@@ -0,0 +1,138 @@
...
+        val increasedCritChance = pw?.additionalCritChanceMutilate() ?: 0.0
+
+        val lethality = sp.character.klass.talents[Lethality.name] as Lethality?
+        val critDmgMultiplier = lethality?.critDamageMultiplier() ?: 1.0
+
+        val item = sp.character.gear.mainHand
+        val damageRoll = Melee.baseDamageRoll(sp, item) + bonusDamage
+        val result = Melee.attackRoll(sp, damageRoll, item, isWhiteDmg = false, bonusCritChance = increasedCritChance, abilityAdditionalCritDamageMultiplier = critDmgMultiplier)
+
+        if(result.second != Event.Result.MISS && result.second != Event.Result.DODGE) {
```
secretbis: This should account for parry

src/commonMain/kotlin/character/classes/rogue/abilities/Mutilate.kt
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    fun poisonsPresentOnTarget(sp: SimParticipant): Boolean {
+        val dpDot = sp.sim.target.debuffs[DeadlyPoisonDot.name]
+        val wpDot = sp.sim.target.debuffs[character.classes.rogue.debuffs.WoundPoison.name]
+        return dpDot != null || wpDot != null
+    }
+
+    override fun cast(sp: SimParticipant) {
+        val poisonBonus = if(poisonsPresentOnTarget(sp)) { 1.5 } else { 1.0 }
```
secretbis: The poison bonus should apply to the full attack, not just the bonus damage

src/commonMain/kotlin/character/classes/rogue/abilities/Rupture.kt
```diff
@@ -0,0 +1,66 @@
...
+        val wouldBeDebuff = character.classes.rogue.debuffs.RuptureDot(sp, wouldBeCombopoints)
+        val canAddDebuff = sp.shouldApplyBuff(wouldBeDebuff, sp.buffs)
+        
+        return canAddDebuff && super.available(sp)
+    }
+
+    override fun cast(sp: SimParticipant) {
+        val result = Melee.attackRoll(sp, 0.0, null, isWhiteDmg = false, noDodgeAllowed = noDodgeAllowed(sp))
+
+        val event = Event(
```
secretbis, line 40: We can omit the damage event for Rupture - applying the dot and firing on-hit procs should be all we need

src/commonMain/kotlin/character/classes/rogue/abilities/Rupture.kt
```diff
@@ -0,0 +1,66 @@
...
+            eventType = Event.Type.DAMAGE,
+            damageType = Constants.DamageType.PHYSICAL,
+            abilityName = name,
+            comboPointsSpent = consumedComboPoints,
+            amount = 0.0,
+            result = result.second,
+        )
+        sp.logEvent(event)
+
+        if(result.second != Event.Result.MISS && result.second != Event.Result.DODGE) {
```
secretbis: This should account for parry

src/commonMain/kotlin/character/classes/rogue/abilities/Shiv.kt
```diff
@@ -0,0 +1,73 @@
...
+
+        val surprise = sp.character.klass.talents[SurpriseAttacks.name] as SurpriseAttacks?
+        val increasedDamagePercent = surprise?.damageIncreasePercent() ?: 0.0
+        val dmgMultiplier = 1 + (increasedDamagePercent / 100.0).coerceAtLeast(0.0)
+
+        val item = sp.character.gear.offHand
+        val damageRoll = Melee.baseDamageRoll(sp, item, isNormalized = true) * dmgMultiplier
+        val result = Melee.attackRoll(sp, damageRoll, item, isWhiteDmg = false, abilityAdditionalCritDamageMultiplier = critDmgMultiplier)
+
+        if(result.second != Event.Result.MISS && result.second != Event.Result.DODGE) {
```
secretbis: This should account for parry

src/commonMain/kotlin/character/classes/rogue/abilities/Shiv.kt
```diff
@@ -0,0 +1,73 @@
...
+            // TODO: unsure if best way to do this
```
secretbis: This is exactly how I'd do it!

src/commonMain/kotlin/character/classes/rogue/abilities/SinisterStrike.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    val bonusDamage = 98.0
+    override fun cast(sp: SimParticipant) {
+        val lethality = sp.character.klass.talents[Lethality.name] as Lethality?
+        val critDmgMultiplier = lethality?.critDamageMultiplier() ?: 1.0
+
+        var increasedDamagePercent = 0.0
+        
+        val aggression = sp.character.klass.talents[Aggression.name] as Aggression?
+        increasedDamagePercent += aggression?.damageIncreasePercent() ?: 0.0
```
secretbis, line 36: Do these stack additively instead of multiplicatively?

src/commonMain/kotlin/character/classes/rogue/abilities/SinisterStrike.kt
```diff
@@ -0,0 +1,78 @@
...
+
+
+        // TODO:
+        // unsure if dmg calculation is correct with the multiplier
+        // from testing this seems to do slightly too much dmg.
+        val item = sp.character.gear.mainHand
+        val damageRoll = (Melee.baseDamageRoll(sp, item, isNormalized = true) + bonusDamage) * dmgMultiplier
+        val result = Melee.attackRoll(sp, damageRoll, item, isWhiteDmg = false, abilityAdditionalCritDamageMultiplier = critDmgMultiplier)
+        
+        if(result.second != Event.Result.MISS && result.second != Event.Result.DODGE) {
```
secretbis: This should account for parry

src/commonMain/kotlin/character/classes/rogue/abilities/SliceAndDice.kt
```diff
@@ -0,0 +1,44 @@
...
+        val canAddBuff = sp.shouldApplyBuff(wouldBeBuff, sp.buffs)
+        
+        return canAddBuff && super.available(sp)
+    }
+
+    override fun cast(sp: SimParticipant) {
+        sp.addBuff(character.classes.rogue.buffs.SliceAndDice(sp, consumedComboPoints))
+        
+        val event = Event(
+            eventType = Event.Type.SPELL_CAST,
```
secretbis, line 40: This would create a duplicate cast event with the default behavior in `SimParticipant`

src/commonMain/kotlin/character/classes/rogue/buffs/Poison.kt
```diff
@@ -0,0 +1,24 @@
...
+abstract class Poison(sourceItem: Item) : TempEnchant(sourceItem) {
+
+    override fun modifyStats(sp: SimParticipant): Stats? {
+        sourceItems.first().tempEnchant = this
+        return null
+    }
+
+    public var poisonAbility: Ability? = null
+
+    fun testlol() {
```
secretbis: We can probably remove this :)

src/commonMain/kotlin/character/classes/rogue/buffs/SliceAndDice.kt
```diff
@@ -0,0 +1,51 @@
...
+
+class SliceAndDice(sp: SimParticipant, consumedComboPoints: Int) : Buff() {
+    companion object {
+        const val name = "Slice and Dice"
+    }
+
+    override val name: String = "Slice and Dice"        
+    override val durationMs: Int = getDurationForCombopoints(sp, consumedComboPoints)
+
+    // can only replace when the new buff has a longer remaining duration
```
secretbis, line 22: Wouldn't a higher combo point SnD also replace?  We might be able to omit this and just keep the `available()` check in the Ability version of this.

src/commonMain/kotlin/character/classes/rogue/debuffs/DeadlyPoisonDot.kt
```diff
@@ -0,0 +1,61 @@
...
+class DeadlyPoisonDot(owner: SimParticipant) : Debuff(owner) {
+    companion object {
+        const val name = "Deadly Poison (DoT)"
+    }
+
+    override val name: String = Companion.name
+    override val durationMs: Int = 12000
+    override val tickDeltaMs: Int = 3000
+    val totalTicks = durationMs / tickDeltaMs
+    override val maxStacks: Int = 4
```
secretbis: Deadly stacks up to 5, right?

src/commonMain/kotlin/character/classes/rogue/debuffs/ExposeArmor.kt
```diff
@@ -0,0 +1,47 @@
...
+    companion object {
+        const val name = "Expose Armor"
+    }
+
+    override val name: String = "Expose Armor"
+    override val durationMs: Int = 30000
+
+    val armorReduce = getArmorReduceForCombopoints(owner, consumedComboPoints)
+
+    override val mutex: List<Mutex> = listOf(Mutex.DEBUFF_EXPOSE_ARMOR)
```
secretbis, line 21: This should use `Mutex.DEBUFF_MAJOR_ARMOR` so it properly mutexes with Sunder Armor when part of a player rotation.  Otherwise, this would stack with a raid-buff configured Sunder debuff (or a second raid-buff configured EA).

src/commonMain/kotlin/character/classes/rogue/debuffs/RuptureDot.kt
```diff
@@ -0,0 +1,87 @@
...
+        const val name = "Rupture (DoT)"
+    }
+
+    override val name: String = Companion.name
+    override val durationMs: Int = getDurationForCombopoints(owner, consumedComboPoints)
+    override val tickDeltaMs: Int = 2000
+    
+    val dmgPerTick = getDamageForCombopoints(owner, consumedComboPoints) / (durationMs / tickDeltaMs)
+
+    // can only replace if it has higher dmg
```
secretbis, line 22: Would this also be properly accounted for in `available()` in the Ability version of this?

src/commonMain/kotlin/mechanics/Melee.kt
```diff
@@ -178,14 +182,40 @@ object Melee {
         return Random.nextDouble(min, max) + apToDamage(sp, totalAp, item, isNormalized)
     }
 
+    fun baseDamageRollPure(minDmg: Double, maxDmg: Double): Double {
```
secretbis, line 185: Possibly worth moving this to `General`, as it's not really related to anything melee-specific.  Very minor.

secretbis (changes requested):
@NoBetaBoredom Outstanding and excellent!  My apologies this review took so long.  

Please do feel free to add the preset as well - it can go next to the others.  I made an Assassination preset to test some things, let me know if you'd like me to add that here as well.  

There are still a few todo items until this can be used fully:
1. Hooking the presets/specs into the UI
1. Linking up EP calc and rankings to said presets/specs

I am happy to help with any of the above, let me know!

src/commonMain/kotlin/character/classes/rogue/abilities/Eviscerate.kt
```diff
@@ -0,0 +1,78 @@
...
+        const val name = "Eviscerate"
+    }
+
+    override val id: Int = 26865
+    override val name: String = Companion.name
+
+    override fun gcdMs(sp: SimParticipant): Int = sp.physicalGcd().toInt()
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 35.0
```
secretbis: This should account for the Assassination 4-set bonus (here's an [example](https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/character/classes/shaman/abilities/LightningBolt.kt#L89))

src/commonMain/kotlin/character/classes/rogue/abilities/Envenom.kt
```diff
@@ -0,0 +1,106 @@
...
+        const val name = "Envenom"
+    }
+
+    override val id: Int = 32684
+    override val name: String = Companion.name
+
+    override fun gcdMs(sp: SimParticipant): Int = sp.physicalGcd().toInt()
+
+    override fun resourceType(sp: SimParticipant): Resource.Type = Resource.Type.ENERGY
+    override fun resourceCost(sp: SimParticipant): Double = 35.0
```
secretbis: This should account for the Assassination 4-set bonus (here's an [example](https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/character/classes/shaman/abilities/LightningBolt.kt#L89))

src/commonMain/kotlin/character/classes/rogue/buffs/SliceAndDice.kt
```diff
@@ -0,0 +1,51 @@
...
+            3 -> 15000
+            4 -> 18000
+            5 -> 21000
+            else -> 0
+        }
+
+        val improved = sp.character.klass.talents[ImprovedSliceAndDice.name] as ImprovedSliceAndDice?
+        val multiplier = improved?.durationMultiplier() ?: 1.0
+        
+        return (baseDuration * multiplier).toInt()
```
secretbis, line 49: This should account for the T4 Netherblade 2-set bonus

src/commonMain/kotlin/character/classes/rogue/abilities/Eviscerate.kt
```diff
@@ -0,0 +1,78 @@
...
+            Event.Result.BLOCKED_CRIT -> listOf(Proc.Trigger.MELEE_YELLOW_CRIT, Proc.Trigger.PHYSICAL_DAMAGE)
+            else -> null
+        }
+
+        fireProcAsFinisher(sp, triggerTypes, null, event)
+    }
+
+    fun damage(sp: SimParticipant, cps: Int, dmgMultiplier: Double): Pair<Double, Double> {
+        val modifier = sp.attackPower() * (cps * 0.03)
+        return when(consumedComboPoints){
```
secretbis, line 78: This should account for the T5 Deathmantle 2-set bonus

src/commonMain/kotlin/character/classes/rogue/abilities/Envenom.kt
```diff
@@ -0,0 +1,106 @@
...
+    override fun available(sp: SimParticipant): Boolean {
+        // there has to be at least one deadly poison debuff available
+        val dosesAvailable = sp.sim.target.debuffState[DeadlyPoisonDot.name] != null
+        
+        return dosesAvailable && super.available(sp)
+    }
+
+    override fun cast(sp: SimParticipant) {
+        val consumedPoisonDoses = min(maxPoisonDosesAvailable(sp), consumedComboPoints)
+        val damage: Double = when(consumedPoisonDoses){
```
secretbis, line 49: This should account for the T5 Deathmantle 2-set bonus

src/commonMain/kotlin/character/classes/rogue/buffs/SliceAndDice.kt
```diff
@@ -0,0 +1,51 @@
...
+    }
+    
+    override fun modifyStats(sp: SimParticipant): Stats {  
+        return Stats(
+            physicalHasteMultiplier = 1.3
```
secretbis: This should account for the T6 Slayer's 2-set bonus

src/commonMain/kotlin/character/classes/rogue/abilities/Backstab.kt
```diff
@@ -0,0 +1,86 @@
...
+        val critDmgMultiplier = lethality?.critDamageMultiplier() ?: 1.0
+
+        val aggression = sp.character.klass.talents[Aggression.name] as ImprovedEviscerate?
+        var increasedDamagePercent = aggression?.damageIncreasePercent() ?: 0.0
+        val surprise = sp.character.klass.talents[SurpriseAttacks.name] as SurpriseAttacks?
+        increasedDamagePercent += surprise?.damageIncreasePercent() ?: 0.0
+        val opportunity = sp.character.klass.talents[Opportunity.name] as Opportunity?
+        increasedDamagePercent += opportunity?.damageIncreasePercent() ?: 0.0
+        
+        val dmgMultiplier = 1 + (increasedDamagePercent / 100.0).coerceAtLeast(0.0)
```
secretbis, line 52: This should account for the T6 Slayer's 4-set bonus

src/commonMain/kotlin/character/classes/rogue/abilities/SinisterStrike.kt
```diff
@@ -0,0 +1,78 @@
...
+        val critDmgMultiplier = lethality?.critDamageMultiplier() ?: 1.0
+
+        var increasedDamagePercent = 0.0
+        
+        val aggression = sp.character.klass.talents[Aggression.name] as Aggression?
+        increasedDamagePercent += aggression?.damageIncreasePercent() ?: 0.0
+        val surprise = sp.character.klass.talents[SurpriseAttacks.name] as SurpriseAttacks?
+        increasedDamagePercent += surprise?.damageIncreasePercent() ?: 0.0
+        
+        val dmgMultiplier = 1 + (increasedDamagePercent / 100.0).coerceAtLeast(0.0)
```
secretbis, line 42: This should account for the T6 Slayer's 4-set bonus

src/commonMain/kotlin/character/classes/rogue/abilities/Mutilate.kt
```diff
@@ -0,0 +1,138 @@
...
+
+        val pw = sp.character.klass.talents[PuncturingWounds.name] as PuncturingWounds?
+        val increasedCritChance = pw?.additionalCritChanceMutilate() ?: 0.0
+
+        val lethality = sp.character.klass.talents[Lethality.name] as Lethality?
+        val critDmgMultiplier = lethality?.critDamageMultiplier() ?: 1.0
+
+        val item = sp.character.gear.mainHand
+        val damageRoll = Melee.baseDamageRoll(sp, item) + bonusDamage
+        val result = Melee.attackRoll(sp, damageRoll, item, isWhiteDmg = false, bonusCritChance = increasedCritChance, abilityAdditionalCritDamageMultiplier = critDmgMultiplier)
```
secretbis, line 66: This should account for the T6 Slayer's 4-set bonus

src/commonMain/kotlin/character/classes/rogue/abilities/Mutilate.kt
```diff
@@ -0,0 +1,138 @@
...
+    fun castOffhand(sp: SimParticipant, bonusDamage: Double) {
+        val opportunity = sp.character.klass.talents[Opportunity.name] as Opportunity?
+        val increasedDamagePercent = opportunity?.damageIncreasePercent() ?: 0.0
+        
+        val dmgMultiplier = 1 + (increasedDamagePercent / 100.0).coerceAtLeast(0.0)
+
+        // TODO: not sure if correct with the multiplier
+        val item = sp.character.gear.offHand
+        val damageRoll = Melee.baseDamageRoll(sp, item) + bonusDamage * dmgMultiplier
+        val result = Melee.attackRoll(sp, damageRoll, item, isWhiteDmg = false)
```
secretbis: This should account for the T6 Slayer's 4-set bonus

secretbis (changes requested):
Ah, forgot about the tier set bonuses.  The tier set abilities are already implemented, and should just need to be referenced in all the relevant abilities.  Non-ability-specific tier set bonuses should already be functional.

They are all in the `itemsets` package, and should be pretty much ready for use.  

Here's one example on how to reference an item set buff: https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/character/classes/shaman/abilities/LightningBolt.kt#L89

src/commonMain/kotlin/sim/SimParticipant.kt
NoBetaBoredom: [Here's the stats of my troll rogue on beta](https://i.imgur.com/3BTA9hY.png). He has gear on here but that doesn't matter. I also tried this with a naked char. Rogue gets 1% crit per 40 agility

First, you can see that for AP calculation 10 agility (and also strength) are ignored.
If you calculate the crit %, you can see it doesn't add up:

- 510/40 = 12.75%
- subtract the same 10 agi that are ignored and you get 500/40 = 12.5%
- _also_ subtract the 2 agi bonus a troll gets from his race: 498/40 = 12.45%

I'm not entirely sure if this is just a display issue or not, but it looks like race-specific bonuses are ignored for crit % calculation. And 10 of agi/str are ignored completely, easily confirmable by the AP amount you get.

src/commonMain/kotlin/character/classes/rogue/Rogue.kt
NoBetaBoredom: [All melees have actually](https://wowwiki-archive.fandom.com/wiki/Attack_power?oldid=1657524). It's mostly charLevel * 2 or 3. I just went based off the displayed values ingame though. The link has a "-20" because it handles the 10 base agi/str that are subtracted. Maybe I should do that here too instead of in `sp.attackPower().`

src/commonMain/kotlin/character/classes/rogue/abilities/AdrenalineRush.kt
NoBetaBoredom: Just went off what you said in my last PR, if this happens it's probably a user error in the preset file.

src/commonMain/kotlin/character/classes/rogue/abilities/Backstab.kt
NoBetaBoredom: Catch all comment here for this and the Opportunity thing:
The thing is I don't know whether its additive or multiplicative for the different abilities, so I went with the one that can be converted to both if needed. Information online seems to be sparse or even completely wrong in some cases (like Sinister Strike). I guess I could just have both methods implemented in the respective Talents.

src/commonMain/kotlin/character/classes/rogue/abilities/DeadlyPoison.kt
NoBetaBoredom: I just forgot to remove this comment, it actually only applies to Instant Poison and I already use `Spell.attackRoll` there with a 100% hitchance bonus to get around the spellhit part (since if it procs, the poison is guaranteed to hit). But there's still no partial resists 0/25/50/75 implemented in the sim so it's not doing much right now.

src/commonMain/kotlin/character/classes/rogue/abilities/Envenom.kt
NoBetaBoredom: `Buff.currentStacks` starts at 0, so 0 `currentStacks` means 1 dose available.

src/commonMain/kotlin/character/classes/rogue/debuffs/DeadlyPoisonDot.kt
NoBetaBoredom: Since `currentStacks` starts at 0, a `maxStacks` of 4 means 5 stacks are applied. currentStacks in `Buff` should probably start at 1 for better readability.

src/commonMain/kotlin/character/classes/rogue/abilities/SinisterStrike.kt
NoBetaBoredom: According to [the old WoW wiki](https://wowwiki-archive.fandom.com/wiki/Sinister_Strike_formulas) they are multiplicative, but in the same wiki they also apply the crit bonus on the whole crit instead of only to the crit bonus dmg and other weird things. Using these formulas does way too much dmg compared to playtesting, so I take anything on there with a huge grain of salt.

From playtesting, this additive approach is almost perfect, it just does a tiny bit too much damage right now (my ingame minhit after 100s of casts are _below_ the minhit of the sim, but just slightly, around 1%)

src/commonMain/kotlin/character/classes/rogue/buffs/SliceAndDice.kt
NoBetaBoredom: Well a higher combo point SnD has a longer duration by definition, so yes. But you can replace a 5 CP SnD that has 5s left with a 1 CP SnD that has a 9s duration, so we need this.

The check in the Ability version uses this Mutex check in the end anyways, it's just there because we have to find out _before_ subtracting resources if we are allowed to replace the buff or not, or we would have to deal with refunding resources (and wasted time in the Rotation) after we find out.

src/commonMain/kotlin/character/classes/rogue/debuffs/RuptureDot.kt
NoBetaBoredom: Like in SnD, `sp.shouldApplyBuff()` is the actual function that checks the mutex, and `available()` just creates a test-debuff and checks with `sp.shouldApplyBuff()` if it is allowed to apply for resource-handling purposes.

NoBetaBoredom:
Will do the changes later today or tomorrow!

> There are still a few todo items until this can be used fully:
> 
> 1. Hooking the presets/specs into the UI
> 2. Linking up EP calc and rankings to said presets/specs
> 
> I am happy to help with any of the above, let me know!

It would be great if you could do both of these changes (and add the presets), so I don't have to dig into the UI stuff.
Another change that has to be made for the UI is the multiple resource handling I did in `Main.kt`, as `SimStats.resourceUsageByAbility()` and `SimStats.resourceUsage()` now take the resourceType as an argument. I also added minHit/maxHit and minCrit/maxCrit, unsure if that needs to be accounted for too.

src/commonMain/kotlin/character/classes/rogue/abilities/Backstab.kt
NoBetaBoredom: The dangers of copy&paste

src/commonMain/kotlin/character/classes/rogue/debuffs/DeadlyPoisonDot.kt
NoBetaBoredom: Actually nevermind, I got confused by how a buff is applied in `SimParticipant.addDebuff()`. `Buff.reset()` sets it to 0, but since `Buff.refresh()` is called on every buff, it's set to 1 always. Still confusing because for non-stacking buffs/debuffs the `currentStacks` will always be 0 instead of a more logical 1 (also means "AvgStacks" shows 0 in the stats).

src/commonMain/kotlin/character/classes/rogue/abilities/Envenom.kt
NoBetaBoredom: Actually nevermind, you are correct, see my comment below for `DeadlyPoisonDot.kt`.

src/commonMain/kotlin/character/classes/rogue/talents/VilePoisons.kt
NoBetaBoredom: I mean, yes, but it's not hurting and more verbose?

src/commonMain/kotlin/character/classes/rogue/abilities/Envenom.kt
NoBetaBoredom: From a quick google search it looks like its affected by nature resistance. It still uses melee hit/crit chances though so will need some modification of `Spell.attackRoll()`

src/commonMain/kotlin/character/classes/rogue/abilities/Rupture.kt
NoBetaBoredom: Rupture is a finisher and the event passes the comboPointsSpent to procs that need that information, so I just commented out the actual logging with `sp.logEvent()`

src/commonMain/kotlin/character/classes/rogue/abilities/SliceAndDice.kt
NoBetaBoredom: The event is only used to carry the comboPointsSpent to Procs that need it, and an `eventType` is mandatory. It's also not logged with `sp.logEvent()`. SPELL_CAST seemed the most fitting, maybe I can change it to `PROC`?

src/commonMain/kotlin/character/classes/rogue/talents/DirtyDeeds.kt
NoBetaBoredom: Reworked it with it's own state now. The talent just applies a `Buff`, and this `Buff` applies the actual damage-bonus-buff once the targets health is below 35% (which is checked in the `Proc` that basically procs off everything). This should work now. There might still be issues because `yellowDamageMultiplier` might apply to damage that is not supposed to be affected by Dirty Deeds.

src/commonMain/kotlin/character/classes/rogue/talents/DirtyDeeds.kt
NoBetaBoredom: Went for the rework instead, see first comment. Can still do this one if not satisifed.

src/commonMain/kotlin/data/itemsets/Deathmantle.kt
```diff
@@ -23,14 +26,12 @@ class Deathmantle : ItemSet() {
...
-                Trigger.ROGUE_CAST_EVISCERATE,
-                Trigger.ROGUE_CAST_ENVENOM,
-                Trigger.ROGUE_CAST_RUPTURE,
-                Trigger.ROGUE_CAST_SLICE_AND_DICE,
+                Trigger.ROGUE_CAST_FINISHER
             )
             override val type: Type = Type.STATIC
 
+            // TODO: is this also consumed on misses/dodges/parry etc? otherwise we have to check here for a hit
+            // if it does consume on a miss/etc. we need to check interactions with the "Quick Recovery" talent.
```
NoBetaBoredom, line 34: See comment. If it does consume on misses, "Quick Recovery" will refund energy when it shouldn't with the current implementation.

src/commonMain/kotlin/data/itemsets/Deathmantle.kt
```diff
@@ -64,6 +65,7 @@ class Deathmantle : ItemSet() {
             override val type: Type = Type.PPM
             override val ppm: Double = 1.0
 
+            // TODO: does this also happen on misses/dodges/parry etc? otherwise we have to check here for a hit
```
NoBetaBoredom, line 68: See comment.

src/commonMain/kotlin/data/itemsets/Netherblade.kt
```diff
@@ -26,15 +29,12 @@ class Netherblade : ItemSet() {
...
-                Trigger.ROGUE_CAST_EVISCERATE,
-                Trigger.ROGUE_CAST_ENVENOM,
-                Trigger.ROGUE_CAST_RUPTURE,
-                Trigger.ROGUE_CAST_SLICE_AND_DICE,
+                Trigger.ROGUE_CAST_FINISHER
             )
             override val type: Type = Type.PERCENT
             override fun percentChance(sp: SimParticipant): Double = 15.0
 
+            // TODO: does this also happen on misses/dodges/parry etc? otherwise we have to check here for a hit
```
NoBetaBoredom, line 37: See comment.

NoBetaBoredom:
Ok, done with all the changes that I marked resolved. Waiting for feedback on the open ones that I commented on.

There's still uncertainty about talent/damage calculation interactions for many abilities (additive/multiplicative), there's just no information available online and some of it is hard or very time intensive to test ingame, especially if it needs to be statistically accurate.

secretbis:
@NoBetaBoredom This is an awesome start, thank you again!  We can tweak as more is discovered.

PR #90, opened 2021-08-30
Add Ring of the Fallen God and Band of the Inevitable from Vanilla

CyberQuackGames:
Adds missing items:
* Ring of the Fallen God (AQ40 ring)
* Band of the Inevitable (Naxxramas ring)

Marked items as non-autogenerated. I wasn't sure where the items.json is generated from, so stayed away from touching that. Icons were already available from the looks of it.


secretbis (changes requested):
Thanks for the contribution!  

In order for these to be actually available, they need to be added to the main item index: https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/data/Items.kt

Also, the `data.items` package is meant for autogenerated items only, as it will be nuked by an item regeneration.  This happens from time to time if items need additional features and/or data, or are refactored for some reason.  

Could you move both of these files to the `data.itemscustom` package?  You'll see several other Vanilla items in that same folder.  Then, add both of them to `data/Items.kt` right about here: https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/data/Items.kt#L15, in the same manner as those already present.  

Edit: `items.json` is just a bulk export from a WoW item database I have, picking out some desired properties from the game data.  It does not need to be updated in this case :)

src/commonMain/kotlin/data/items/BandOfTheInevitable.kt
```diff
@@ -0,0 +1,63 @@
...
+
+  public override var stats: Stats = Stats()
+
+  public override var sockets: Array<Socket> = arrayOf()
+
+  public override var socketBonus: SocketBonus? = null
+
+  public override var phase: Int = 1
+
+  public override val buffs: List<Buff> by lazy {
```
secretbis: The stat buffs provided by this item can be implemented directly.  

```
// Implement the stats directly in this property
public override var stats: Stats = Stats(
  spellHitRating = 8.0,
  spellDamage = 36
)

// Remove game-data strings in this property
public override val buffs: List<Buff> = listOf()
```

src/commonMain/kotlin/data/items/RingOfTheFallenGod.kt
```diff
@@ -0,0 +1,66 @@
...
+      intellect = 6
+      )
+
+  public override var sockets: Array<Socket> = arrayOf()
+
+  public override var socketBonus: SocketBonus? = null
+
+  public override var phase: Int = 1
+
+  public override val buffs: List<Buff> by lazy {
```
secretbis: The stat buffs provided by this item can be implemented directly.  

```
// Implement the stats directly in this property
public override var stats: Stats = Stats(
  stamina = 5,
  intellect = 6,
  spellHitRating = 8.0,
  spellDamage = 37
)

// Remove game-data strings in this property
public override val buffs: List<Buff> = listOf()
```

secretbis (changes requested):
Ah, noticed one more thing - it would be simpler to implement the stats directly than going through the lazy game-data-string parsing mechanism.  That just exists to make item generation simpler by offloading some logic.

When creating items by hand, implementing the `Stats` object directly adds more clarity.  The `buffs` list is normally used for items with complex procs, though the game engine treats all these things relatively the same.

CyberQuackGames:
I updated with suggested changes here. Thanks for the detailed break down. Been diving into the project a bit to see how it works and experiment with builds.

secretbis (approved):
Great work, thanks for the changes!

PR #92, opened 2021-09-05
Shadow Priest

CyberQuackGames:
### Updates

* Adds Priest base class
* Adds Shadow Priest related abilities / talents
* Adds Shadow Priest specs, gear, and rotation
* Adds `SPELL_START_CAST` proc event (Needed for sudo channeled spells)
* Adds `ARCANE_DAMAGE_PERIODIC` proc event (Needed for Starshards)
* Adds damage type (E.g spell schools) to auto attack base where required (Shadowfiend)
* Adds option to ignore crit chance for spell damage calculations (Dots / Channeled related)
* Adds `canCrit`, `canResist`, and `bonusDamageMultiplier` to `attackRoll`s for spells. This enables dot snap shot damage calculations as they are needed and provides additional bonus damage multipliers for buffs like item sets.

#### Dot Snapshotting

Currently dots do not snap shot (Re-calculting spell power each tick). In order to simulate
shadow priest a bit more accurately, dots calculate the base spell damage up front and pass 
this information to the dot debuff for tick damage. This pattern is consistent for all dots
with Shadow Priest (Shadow Word Pain, Vampiric Touch, and Mind Flay (More below)).

Second, when dot damage is calculated off the snap shot, the damage should still consider
enemy "damage taken multipliers" into effect in real time. If for example the enemy has
Shadow Weaving at 5 stacks (10% shadow damage) the overall snapshot does not take this into account,
but the individual tick damage will calculate this every tick. These updates considering
target damage taken would normally apply to many other debuffs such as Misery, Curse of Elements,
Improved Shadow Bolt, etc, however most other buffs are considered "Raid buffs" currently.
These raid buffs could be moved to raid target debuffs, but the damage output would be the same
regardless.

Lastly in the future when haste becomes a factor for dots, the update to these dots
should be trivial going forward as haste can factor into dot duration which was needed
to simulate Mind Flay.

#### Channeled Spell Concept

Shadow Priests core filler ability Mind Flay is a channeled spell in game. At the moment no other rotations
use channeled spells for dps output, although spells like Arcane Missiles and Drain Life do exist.

In order to simulate this spell there was a few items to work around. From a pure technical standpoint 
we can consider Mind Flay a ahead of it's time "haste driven dot" where the user can't move or cast any
other spell. To this point, shadow priests needed the new `SPELL_START_CAST` proc in order to simulate
"canceling" the Mind Flay "dot". In the sim currently we don't have to worry about movement so this factor
is negated.

From a logs stand point all channeled spells are actually considered
a periodic damage dot:

* [Mind Flay](https://tbc.wowhead.com/spell=25387/mind-flay)
* [Arcane Missles](https://tbc.wowhead.com/spell=38704/arcane-missiles)
* [Drain Life](https://tbc.wowhead.com/spell=27220/drain-life)

The only main difference between these "dots" and regular dots such was Shadow Word: Pain or Vampiric Touch
is that they are all affected by haste (Similar to how Wrath introduces haste to regular dots). The implementation
across all dots allows for adding haste in later with ease, but only Mind Flay's "dot" is affected by haste for now.

In terms of rotation there are 3 possible outcomes (assuming the initial cast hit):

* Mind Flay 2 ticks - Mind Flay is cast with the intent to stop at 2 ticks out of 3. Without haste this is 2 seconds after cast
* Mind Flay 3 ticks - Mind Flay is cast with the intent to go to full duration (3 seconds / 3 ticks)
* Mind Flay with interrupt - Mind flay is cast with the intent to be interrupted immediately after GCD is up. Without haste this results
in a 0.5 second waste, however it's realistic to assume it's possible to interrupt in some scenarios so the rotation allows for this
by omitting the check for Mind Flay (DoT) being present in which the rotation will choose the next spell as soon as the GCD permits. In
most scenarios Mind Flay shouldn't be interrupted in this way (default on all spells is to not interrupt in the presets).

#### Presets

Shadow priest rotation is probably the most complex one when considering dot uptime and Mind Flay filler priority. Luckily we are only worrying about one target in this sim. Generally this is to keep Vampiric Touch dot up, keep Shadow Word: Pain dot up, Mind Blast on cd, Shadow Word: Death on cd, Mind Flay.

The only other major thing to consider is Shadow Priests do not want to be "casting" Mind Flay when any higher priority spell should be cast or start to cast. As it stands the criteria in the rotations allows for this by specifying each spell with mind flay 2/3 tick variations and ensuring we aren't casting Mind Flay by checking if the dot is present. Since Vampiric Touch isn't on a CD we want to start casting it around the time it takes to full cast it when the dot
would expire (1.5 seconds without haste).

#### Future Sim Ideas For Shadow Priest

* DoT abstraction for all classes (For use in regular DoTs and channeled "DoTs") to help with snap shots and channeled use cases. - Note: I plan on doing this, but didn't want to include it in this large PR already.
* Shadow priest rotation (and I would imagine many others) changes based on haste. The ability to key off run time variables with something like `$CURRENT_GCD * 2` or even `$VAMPIRIC_TOUCH_CAST_TIME * 2` would go a long ways
in optimizing for haste with buffs and gear later.
* Shadow priest shine a lot more in multi-target (e.g Spell Cleave) fights, having a add cleave scenario would be awesome.
* Some itemization for enchant based gear (`of Shadow Wrath`) is missing which is fairly nice for Shadow Priest sims.

Shout out to @erladion as well for some of the initial setup to get this started.


ui/src/ep/data/ep_all.json
```diff
@@ -513,20 +546,6 @@
...
-                "meleeCritRating": 2.27,
-                "physicalHitRating": 1.82,
-                "physicalHasteRating": 2.28,
-                "expertiseRating": 0.0,
-                "armorPen": 0.34,
-                "redSocket": 20.8,
-                "yellowSocket": 22.7,
-                "blueSocket": 10.4,
-                "metaSocket": 100.0
-            },
```
CyberQuackGames, line 529: Looks like the calc auto organized this by name

src/commonMain/kotlin/character/Proc.kt
```diff
@@ -51,6 +51,8 @@ abstract class Proc {
         SPELL_HIT,
         SPELL_CRIT,
         SPELL_RESIST,
+        SPELL_START_CAST,
+        SPELL_CAST,
```
secretbis: I had removed the `SPELL_CAST` Trigger in a recent commit, since it was easy to make some timing mistakes with on-hit procs.  The sim originally fired it as a cast-complete event, after the actual spell cast, but it ends up being generally redundant with the spell damage events.  

One example of an issue caused with that was Mage Arcane Concentration (aka Clearcasting).  The AC buff was configured to consume on `SPELL_CAST`, and refund the cost of the spell.  Since the `SPELL_CAST` event was fired immediately on the same tick after the `SPELL_HIT/CRIT` that triggered the proc, it would be immediately consumed by the spell that triggered it.  

This was refactored to trigger on a trio of `SPELL_CRIT`, `SPELL_HIT`, and `SPELL_RESIST` to fit within the usual model of abilities triggering on "damaging spells".  With `SPELL_CAST`, it wouldn't distinguish between a damaging and non-damaging spell.  

Inner Focus below, for example, would be cast and immediately consumed by the subsequent `SPELL_CAST` event.

TLDR: `SPELL_CAST` is a bit dangerous and should be removed (the sim doesn't fire it anymore), let's see if we can be more explicit about spells that listen for these types of events.

src/commonMain/kotlin/character/Stats.kt
```diff
@@ -99,6 +100,10 @@ data class Stats(
     var arcaneDamageMultiplier: Double = 1.0,
     var petDamageMultiplier: Double = 1.0,
 
+    // Spell school damage taken increases
+    var spellDamageTakenMultiplier: Double = 1.0,
```
secretbis, line 104: While I understand the purpose of these new stats from your PR description, I am not sure it will work consistently.  Currently, I only see Shadow Weaving as modifying `shadowDamageTakenMultiplier`, which excludes things like Misery, Curse of the Elements, and Malediction.  

The snapshot implementations already take all those debuffs into account for the initial damage, since they go through the base `Spell` mechanics.  Raid buffs like the above use generic or school multipliers to implement themselves.  

I think a better way to implement snapshotting is a new arg to `Spell.baseDamageRollSingle` and `Spell.baseDamageRoll` which is just a spellpower value snapshot.  The code there would then just use that in place of `sp.spellDamage()`.  

This keeps the model of damage-modification buffs/debuffs consistent, and simplifies the damage calculation in the DoT spells.

For the final resoult, we can add a `canResist` arg to `Spell.attackRoll`, then use that appropriately.  `canCrit=false` in combination with `canResist=false` should have the effect of an automatic hit that takes into account general and school-specific damage modifiers.

Alternatively, we can combine the above two items into a special one-function snapshotted-DoT damage helper, which could simplify the above somewhat.

src/commonMain/kotlin/character/classes/priest/abilities/MindBlast.kt
```diff
@@ -0,0 +1,93 @@
...
+
+        return baseResourceCost * fmMulti
+    }
+    
+    override fun cast(sp: SimParticipant) {
+        val spTalent: ShadowPower? = sp.character.klass.talentInstance(ShadowPower.name)
+        val spCrit = spTalent?.critIncreasePct() ?: 0.0
+        val sfTalent: ShadowFocus? = sp.character.klass.talentInstance(ShadowFocus.name)
+        val sfHit = sfTalent?.shadowHitIncreasePct() ?: 0.0
+        val targetMultiplier = sp.sim.target.stats.getSpellDamageTakenMultiplier(school)
```
secretbis: I'm not sure I like individual abilities computing this - this should be part of base mechanics, and handled more automatically like other generic buffs/debuffs.  

Perhaps we can roll this into `shadowDamageMultiplier` for now?

src/commonMain/kotlin/character/classes/priest/abilities/ShadowWordDeath.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        return baseResourceCost
+    }
+    
+    override fun cast(sp: SimParticipant) {
+        val spTalent: ShadowPower? = sp.character.klass.talentInstance(ShadowPower.name)
+        val spCrit = spTalent?.critIncreasePct() ?: 0.0
+        val sfTalent: ShadowFocus? = sp.character.klass.talentInstance(ShadowFocus.name)
+        val sfHit = sfTalent?.shadowHitIncreasePct() ?: 0.0
+        val targetMultiplier = sp.sim.target.stats.getSpellDamageTakenMultiplier(school)
```
secretbis: See above note regarding rolling this into general mechanics.

src/commonMain/kotlin/character/classes/priest/debuffs/MindFlayDot.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    val school = Constants.DamageType.SHADOW
+
+    val mindFlay = object : Ability() {
+        override val id: Int = 25387
+        override val name: String = Companion.name
+        override fun gcdMs(sp: SimParticipant): Int = 0
+
+        override fun cast(sp: SimParticipant) {
+            val spellMultiplier = sp.stats.getSpellDamageTakenMultiplier(school)
```
secretbis: See above note regarding rolling this into general mechanics.

src/commonMain/kotlin/character/classes/priest/debuffs/ShadowWordPainDot.kt
```diff
@@ -0,0 +1,52 @@
...
+    val baseDamage = damageRoll / baseTickCount
+
+    val ability = object : Ability() {
+        override val id: Int = 25368
+        override val name: String = Companion.name
+
+        override fun gcdMs(sp: SimParticipant): Int = 0
+
+        override fun cast(sp: SimParticipant) {
+            val spellMultiplier = sp.stats.getSpellDamageTakenMultiplier(school)
```
secretbis: See above note regarding rolling this into general mechanics.

src/commonMain/kotlin/character/classes/priest/debuffs/VampiricTouchDot.kt
```diff
@@ -0,0 +1,52 @@
...
+    val baseDamage = damageRoll / baseTickCount
+
+    val ability = object : Ability() {
+        override val id: Int = 34917
+        override val name: String = Companion.name
+
+        override fun gcdMs(sp: SimParticipant): Int = 0
+
+        override fun cast(sp: SimParticipant) {
+            val spellMultiplier = sp.stats.getSpellDamageTakenMultiplier(school)
```
secretbis: See above note regarding rolling this into general mechanics.

src/commonMain/kotlin/character/races/debuffs/DevouringPlagueDot.kt
```diff
@@ -0,0 +1,56 @@
...
+    val baseDamage = damageRoll / baseTickCount
+
+    val dpAbility = object : Ability() {
+        override val id: Int = 25467
+        override val name: String = Companion.name
+
+        override fun gcdMs(sp: SimParticipant): Int = 0
+
+        override fun cast(sp: SimParticipant) {
+            val spellMultiplier = sp.stats.getSpellDamageTakenMultiplier(school)
```
secretbis: See above note regarding rolling this into general mechanics.

src/commonMain/kotlin/character/races/debuffs/StarshardsDot.kt
```diff
@@ -0,0 +1,52 @@
...
+    val baseDamage = damageRoll / baseTickCount
+
+    val starshardsAbility = object : Ability() {
+        override val id: Int = 25446
+        override val name: String = Companion.name
+
+        override fun gcdMs(sp: SimParticipant): Int = 0
+
+        override fun cast(sp: SimParticipant) {
+            val spellMultiplier = sp.stats.getSpellDamageTakenMultiplier(school)
```
secretbis: See above note regarding rolling this into general mechanics.

src/commonMain/kotlin/character/classes/priest/talents/MindFlay.kt
```diff
@@ -0,0 +1,50 @@
...
+        override val name = "${Companion.name} (Static)"
+        override val hidden = true
+        override val durationMs = -1
+
+        // In general only the GCD prevents another spell from being cast for channeled spells
+        // When another spell is cast, mind flay "dot" should stop ticking
+        val interruptProc = object : Proc() {
+            override val triggers: List<Trigger> = listOf(
+                Trigger.SPELL_START_CAST,
+                Trigger.SPELL_CAST
```
secretbis: See above note about `SPELL_CAST` proc removal.

src/commonMain/kotlin/character/races/abilities/DevouringPlague.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    override fun cast(sp: SimParticipant) {
+        val sfTalent: ShadowFocus? = sp.character.klass.talentInstance(ShadowFocus.name)
+        val sfHit = sfTalent?.shadowHitIncreasePct() ?: 0.0
+
+        // snapshot damage on initial cast
+        val damageRoll = Spell.baseDamageRollSingle(sp, baseDamage, school, spellPowerCoeff)
+        val result = Spell.attackRoll(sp, damageRoll, school, isBinary = true, bonusHitChance = sfHit, canCrit = false)
+
+        val event = Event(
```
secretbis, line 51: This is likely a duplicate, since `SPELL_CAST` is logged (but not procced) by the sim after every rotation ability.

src/commonMain/kotlin/character/races/abilities/Starshards.kt
```diff
@@ -0,0 +1,59 @@
...
+    override fun available(sp: SimParticipant): Boolean {
+        return sp.character.race is NightElf && sp.character.klass is Priest && super.available(sp)
+    }
+
+    override fun cast(sp: SimParticipant) {
+        // snapshot damage on initial cast
+        val damageRoll = Spell.baseDamageRollSingle(sp, baseDamage, school, spellPowerCoeff)
+        val result = Spell.attackRoll(sp, damageRoll, school, isBinary = true, canCrit = false)
+
+        val event = Event(
```
secretbis, line 35: This is likely a duplicate, since `SPELL_CAST` is logged (but not procced) by the sim after every rotation ability.

ui/src/presets/samples/priest_shadow_phase2.yml
```diff
@@ -0,0 +1,312 @@
...
+  mainHand:
+    name: Nathrezim Mindblade
+    enchant: Soulfrost
+    tempEnchant: Superior Wizard Oil
+  offHand:
+    name: Orb of the Soul-Eater
+  rangedTotemLibram:
+    name: Wand of the Forgotten Star
+  head:
+    name: Spellstrike Hood
```
secretbis: Some BIS lists I have found wear 4p T5, and keep Soul-Eater's Handwraps.  Thoughts?

src/commonMain/kotlin/character/classes/priest/abilities/MindFlay.kt
```diff
@@ -0,0 +1,81 @@
...
+        val damageRoll = Spell.baseDamageRollSingle(sp, baseDamage, school, spellPowerCoeff)
+        val resultTick = Spell.attackRoll(
+            sp, 
+            damageRoll,
+            school,
+            isBinary = true,
+            bonusHitChance = sfHit,
+            canCrit = false,
+        )
+        val initialCast = Event(
```
secretbis: The sim logs a `SPELL_CAST` event for every rotation item, so this event is likely redundant.

src/commonMain/kotlin/character/classes/priest/abilities/ShadowWordPain.kt
```diff
@@ -0,0 +1,71 @@
...
+        val iswpTicks = iswp?.currentRank ?: 0
+        val sfTalent: ShadowFocus? = sp.character.klass.talentInstance(ShadowFocus.name)
+        val sfHit = sfTalent?.shadowHitIncreasePct() ?: 0.0
+
+        // snapshot damage on initial cast
+        val tickCount = baseDotTickCount + iswpTicks
+        val damageRoll = Spell.baseDamageRollSingle(sp, baseDamage, school, spellPowerCoeff)
+        val result = Spell.attackRoll(sp, damageRoll, school, isBinary = true, bonusHitChance = sfHit, canCrit = false)
+
+        val event = Event(
```
secretbis, line 47: The sim logs a `SPELL_CAST` event for every rotation item, so this event is likely redundant.

src/commonMain/kotlin/character/classes/priest/buffs/InnerFocus.kt
```diff
@@ -0,0 +1,39 @@
...
+        const val name = "Inner Focus"
+    }
+
+    override val id = 14751
+    override val name: String = Companion.name
+    override val durationMs: Int = -1
+
+    fun genCastRemovalProc(ifBuff: InnerFocus) = object : Proc() {
+        override val triggers: List<Trigger> = listOf(
+            Trigger.SPELL_CAST
```
secretbis: See above note about `SPELL_CAST` proc removal.

src/commonMain/kotlin/character/classes/priest/abilities/Shadowfiend.kt
```diff
@@ -0,0 +1,48 @@
...
+    companion object {
+        const val name: String = "Shadowfiend"
+    }
+
+    override val id: Int = 34433
+    override val name: String = Companion.name
+
+    val fiendUpBuff = object : Buff() {
+        override val name: String = "${Shadowfiend.name} (Hidden)"
+        override val durationMs: Int = 15000
```
secretbis: I see the `IncarnateRegalia` import above - should that set bonus be checked here?

src/commonMain/kotlin/character/classes/priest/abilities/MindFlay.kt
```diff
@@ -0,0 +1,81 @@
...
+    override fun available(sp: SimParticipant): Boolean {
+        return super.available(sp) && sp.character.klass.hasTalentRanks(MindFlayTalent.name)
+    }
+
+    override fun cast(sp: SimParticipant) {
+        val sfTalent: ShadowFocus? = sp.character.klass.talentInstance(ShadowFocus.name)
+        val sfHit = sfTalent?.shadowHitIncreasePct() ?: 0.0
+
+        // Overall damage is determined before doing tick damage
+        val damageRoll = Spell.baseDamageRollSingle(sp, baseDamage, school, spellPowerCoeff)
```
secretbis: I wasn't sure that Mind Flay snapshotted, but I made a PTR priest and it does indeed do so!  TIL!

src/commonMain/kotlin/character/classes/priest/debuffs/ShadowWeavingDebuff.kt
```diff
@@ -0,0 +1,20 @@
...
+import sim.SimParticipant
+
+class ShadowWeavingDebuff(sp: SimParticipant) : Debuff(sp){
+    override val id = 15258
+    override val name: String = "Shadow Vulnerability"
+    override val durationMs: Int = 15000
+    override val maxStacks: Int = 5
+
+    override fun modifyStats(sp: SimParticipant) = Stats(
+        shadowDamageTakenMultiplier = 1 + (0.02 * state(sp).currentStacks)
```
secretbis: The [raid version of this](https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/data/abilities/raid/ShadowWeaving.kt) is currently implemented as a buff which modifies `shadowDamageMultiplier`.  

I think it would be good to make this use that same mechanism, if possible.  

Also, other class-specific debuffs that have a raid version have the same name so that having both enabled does not double-dip (buffs are keyed by name).  We should ensure that if a version of this ends up as a "buff" on the player, it should match the name of the raid version.

src/commonMain/kotlin/character/classes/priest/abilities/MindFlay.kt
```diff
@@ -0,0 +1,81 @@
...
+        sp.logEvent(initialCast);
+
+        if (resultTick.first == 0.0){
+            sp.fireProc(listOf(Proc.Trigger.SPELL_RESIST), listOf(), this, initialCast)
+            return;
+        }
+
+        sp.fireProc(listOf(Proc.Trigger.SPELL_HIT), listOf(), this, initialCast)
+
+        sp.sim.target.addDebuff(MindFlayDot(sp, resultTick.first / 3, tickCount));
```
secretbis: See above note on snapshot implementation.

src/commonMain/kotlin/character/classes/priest/abilities/ShadowWordPain.kt
```diff
@@ -0,0 +1,71 @@
...
+        sp.logEvent(event)
+
+        if(result.first == 0.0){
+            sp.fireProc(listOf(Proc.Trigger.SPELL_RESIST), listOf(), this, event)
+            return
+        }
+
+        sp.fireProc(listOf(Proc.Trigger.SPELL_HIT), listOf(), this, event)
+
+        sp.sim.target.addDebuff(ShadowWordPainDot(sp, result.first, tickCount))
```
secretbis: See above note on snapshot implementation.

src/commonMain/kotlin/character/classes/priest/abilities/VampiricTouch.kt
```diff
@@ -0,0 +1,78 @@
...
+
+        if(result.first == 0.0){
+            sp.fireProc(listOf(Proc.Trigger.SPELL_RESIST), listOf(), this, event)
+            return
+        }
+
+        sp.fireProc(listOf(Proc.Trigger.SPELL_HIT), listOf(), this, event)
+
+        sp.addBuff(VampiricTouchBuff())
+        sp.sim.target.addDebuff(VampiricTouchDot(sp, result.first, baseDotTickCount))
```
secretbis: See above note on snapshot implementation.

src/commonMain/kotlin/character/classes/priest/abilities/VampiricTouch.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    override fun cast(sp: SimParticipant) {
+        val sfTalent: ShadowFocus? = sp.character.klass.talentInstance(ShadowFocus.name)
+        val sfHit = sfTalent?.shadowHitIncreasePct() ?: 0.0
+
+        // snapshot damage on initial cast
+        val damageRoll = Spell.baseDamageRollSingle(sp, baseDamage, school, spellPowerCoeff)
+        val result = Spell.attackRoll(sp, damageRoll, school, isBinary = true, bonusHitChance = sfHit, canCrit = false)
+
+        val event = Event(
```
secretbis, line 60: The sim logs a `SPELL_CAST` event for every rotation item, so this event is likely redundant.

src/commonMain/kotlin/character/classes/priest/pet/buffs/ManaLeech.kt
```diff
@@ -0,0 +1,33 @@
+package character.classes.priest.pet.buffs
+
+import character.*
+import character.classes.hunter.talents.FerociousInspiration
```
secretbis: I assume this import is not needed?  Effects like FI are applied to the entire raid, including their pets (even though an SPriest in a hunter group is highly unlikely).

src/commonMain/kotlin/character/races/abilities/Berserking.kt
```diff
@@ -39,7 +39,8 @@ class Berserking : Ability() {
 
         override fun modifyStats(sp: SimParticipant): Stats {
             return Stats(
-                physicalHasteMultiplier = 1.1
+                physicalHasteMultiplier = 1.1,
+                spellHasteRating = 10 * Rating.hastePerPct
```
secretbis: This should probably modify `spellHasteMultiplier`, since Berserking would have a multiplicative effect with other independent sources of haste (like cumulative spell haste rating).

src/commonMain/kotlin/character/races/abilities/DevouringPlague.kt
```diff
@@ -0,0 +1,75 @@
...
+        sp.logEvent(event)
+
+        if(result.first == 0.0){
+            sp.fireProc(listOf(Proc.Trigger.SPELL_RESIST), listOf(), this, event)
+            return
+        }
+
+        sp.fireProc(listOf(Proc.Trigger.SPELL_HIT), listOf(), this, event)
+
+        sp.sim.target.addDebuff(DevouringPlagueDot(sp, result.first, baseDotTickCount))
```
secretbis: See above note on snapshot implementation.

src/commonMain/kotlin/character/races/abilities/Starshards.kt
```diff
@@ -0,0 +1,59 @@
...
+        sp.logEvent(event)
+
+        if(result.first == 0.0){
+            sp.fireProc(listOf(Proc.Trigger.SPELL_RESIST), listOf(), this, event)
+            return
+        }
+
+        sp.fireProc(listOf(Proc.Trigger.SPELL_HIT), listOf(), this, event)
+
+        sp.sim.target.addDebuff(StarshardsDot(sp, result.first, baseDotTickCount))
```
secretbis: See above note on snapshot implementation.

secretbis (changes requested):
Awesome work!  Thanks for getting this up so quickly!

Please let me know what you think about my take on the snapshot implementation + new stats + Shadow Weaving situation!

src/commonMain/kotlin/character/classes/priest/debuffs/MindFlayDot.kt
```diff
@@ -0,0 +1,47 @@
...
+            val event = Event(
+                eventType = EventType.DAMAGE,
+                damageType = school,
+                abilityName = name,
+                amount = damageRoll * spellMultiplier,
```
secretbis: Should the T4 4-piece bonus be accounted for here?

src/commonMain/kotlin/character/classes/priest/abilities/ShadowWordPain.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    override fun cast(sp: SimParticipant) {
+        val iswp: ImprovedShadowWordPain? = sp.character.klass.talentInstance(ImprovedShadowWordPain.name)
+        val iswpTicks = iswp?.currentRank ?: 0
+        val sfTalent: ShadowFocus? = sp.character.klass.talentInstance(ShadowFocus.name)
+        val sfHit = sfTalent?.shadowHitIncreasePct() ?: 0.0
+
+        // snapshot damage on initial cast
+        val tickCount = baseDotTickCount + iswpTicks
```
secretbis: Should the T6 2-piece bonus be accounted for here?

src/commonMain/kotlin/character/classes/priest/abilities/MindBlast.kt
```diff
@@ -0,0 +1,93 @@
...
+        
+        val damageRoll = Spell.baseDamageRoll(
+            sp, 
+            baseDamage.first,
+            baseDamage.second, 
+            school, 
+            spellPowerCoeff,
+        )
+        val result = Spell.attackRoll(sp, damageRoll, school, bonusHitChance = sfHit, bonusCritChance = spCrit )
+        var finalDamage = result.first * targetMultiplier;
```
secretbis: Should the T6 4-piece be accounted for here?

src/commonMain/kotlin/character/Proc.kt
CyberQuackGames: This was probably missed in the rebase. I'll fix it up.

src/commonMain/kotlin/character/Stats.kt
CyberQuackGames: I had thought about this one quite a bit in both trying to simulate the damage (spell / shadow damage taken are real aura values in game, see [example](https://tbc.wowhead.com/spell=15258/shadow-vulnerability) where `Apply Aura: Mod % Damage Taken (Shadow)`) and not expand damage roll methods too much (Since they had too many parameters as a code smell). I agree the individual calculation felt weird for direct damage spells and should use a damage calculator within `Spell` to be consistent across both direct and DoTs going forward.

I'll add the base spell damage / canResist pieces for snap shots and I'll integrate all damage calcs into base / attack roll going forward. In keeping with spell damage taken stat to simulate, did you want me to go ahead and add `spellDamageTakenMultiplier` to Misery and Curse of Elements here and add those as raid debuffs instead so the spells are consistent or should we revisit this later and assume it's a part of base spell damage for now? In terms of damage on dots there isn't any change as it's just a multiplier regardless, but at least for direct damage it would correctly be applied

src/commonMain/kotlin/character/classes/priest/abilities/MindBlast.kt
CyberQuackGames: I'll integrate with base spell damage calcs as mentioned above, this and SW:D felt weird for sure. I'll see how I can integrate into base calcs while not affecting dots as well.

ui/src/presets/samples/priest_shadow_phase2.yml
CyberQuackGames: I went off https://wowtbc.gg/bis-list/shadow-priest/ as a starting point. There is also other sets where engineering / season 2 weapon are taken into account. We are missing "of Shadow Wrath" gear here as well. I'll throw a few together locally and SIM which one comes out higher and report back.

src/commonMain/kotlin/character/classes/priest/debuffs/MindFlayDot.kt
CyberQuackGames: Yeah I didn't implement set bonuses across the board, I'll take a pass at adding these.

src/commonMain/kotlin/character/Stats.kt
secretbis: > did you want me to go ahead and add spellDamageTakenMultiplier to Misery and Curse of Elements here and add those as raid debuffs instead so the spells are consistent or should we revisit this later and assume it's a part of base spell damage for now? 

It's equivalent for single-target, and there's a lot of other work required before multi-target and complete raid sims can be done - I think it's best to punt on that for now.

src/commonMain/kotlin/character/races/abilities/DevouringPlague.kt
CyberQuackGames: I had copied the initial cast events from SiphonLife / Corruption for the dots I implemented before I realized this was done automatically. For the purposes of dot initial application events with Spell Hits / Resists, what should be the correct event type here? I could use `EventType.Damage`, but it feels weird arbitrarily assigning that event when no damage is being done for initial dot application. I think we need a way to grab the spell cast event for passing to `SPELL_HIT` / `SPELL_RESIST` procs in this case.

CyberQuackGames: So something like below in `SimParticipant`
```
// Log cast event
  val castEvent = Event(
      eventType = EventType.SPELL_CAST,
      abilityName = castingRule!!.ability.name,
      target = sim.target
  )

  castingRule!!.ability.beforeCast(this)
  castingRule!!.ability.cast(this, castEvent)
  castingRule!!.ability.afterCast(this)

  logEvent(castEvent)
```
and update the `Ability` to accept an optional cast event:
```
abstract fun cast(sp: SimParticipant, castEvent: Event?)
```

CyberQuackGames: This would actually clean up quite a few spells since we don't need to generate a no-op event for triggering spell based procs, but what ever you think works here.

src/commonMain/kotlin/character/classes/priest/abilities/MindFlay.kt
CyberQuackGames: Yeah channeled spells are just fancy dots it seems with haste affecting them.

src/commonMain/kotlin/character/classes/priest/debuffs/ShadowWeavingDebuff.kt
CyberQuackGames: Good catch to sync these.

src/commonMain/kotlin/character/classes/priest/pet/buffs/ManaLeech.kt
CyberQuackGames: Yep, probably just copied over / missed it from another spell.

src/commonMain/kotlin/character/races/abilities/Berserking.kt
CyberQuackGames: I thought the mechanics for the multiplier would only use existing haste rating, but upon further review it looks to default to base 1.0 as the base for spell haste so will change this.

src/commonMain/kotlin/character/races/abilities/DevouringPlague.kt
secretbis: > For the purposes of dot initial application events with Spell Hits / Resists, what should be the correct event type here?

In order to work with sim stats output, it would currently have to be a `DAMAGE` event - those all get rolled up into the main abilities table.  I'm not sure how useful it would be to have a new event, since it would have to have all the fields a `DAMAGE` event would normally have, anyway, so the stats code can deal with it.

> So something like below in SimParticipant

I'm not super excited about intentionally mutating an argument, or about changing the signature of `cast` - that would require some rework to every ability, due to how Kotlin handles override signatures.  

I think adding the zero `DAMAGE` event is best, since it will fit most naturally into trinkets like Eye of Magtheridon, and will work out of the box with `SimStats` when computing hit% in the output.

secretbis: SimStats context: https://github.com/marisa-ashkandi/tbcsim/blob/main/src/commonMain/kotlin/sim/SimStats.kt#L215

wobbelbonk:
Hey, Im a top 10 TBC Shadow Priest and "guide" writer for the priest discord, I'd love to work with you on making this as good as it could be. If you're interested add me on Discord @Passion#6241

src/commonMain/kotlin/character/Stats.kt
CyberQuackGames: Ok sounds good, I actually just pushed a change to make shadow weaving a raid buff instead to simplify this whole thing. While it was nice having the true damage taken stat, it definitely complicated the calculation more than needed.

src/commonMain/kotlin/character/classes/priest/abilities/MindBlast.kt
CyberQuackGames: Remove these refs in latest push.

src/commonMain/kotlin/character/classes/priest/debuffs/MindFlayDot.kt
CyberQuackGames: Taken out now

src/commonMain/kotlin/character/classes/priest/abilities/VampiricTouch.kt
CyberQuackGames: I converted all Spell Cast to Spell Damage events. They do show up in the ability log output, but as 0 damage spells. It's kinda nice for seeing how many true base line spells hit compared to dot count though. If there is a better event here to use for the triggers I can swap it up.

src/commonMain/kotlin/character/classes/priest/pet/buffs/ManaLeech.kt
CyberQuackGames: Removed erroneous import

src/commonMain/kotlin/character/races/abilities/Berserking.kt
CyberQuackGames: Fixed up.

src/commonMain/kotlin/character/classes/priest/talents/MindFlay.kt
CyberQuackGames: Updated to HIT / CRIT triggers

src/commonMain/kotlin/character/races/abilities/DevouringPlague.kt
CyberQuackGames: Nice I went with `DAMAGE` event so that works, as I mentioned above it's nice seeing the output in the ability log as well. There are a few other places which have `SPELL_CAST` remaining with other classes, but I can put up a fix PR later for those if no one else gets to em.

src/commonMain/kotlin/character/classes/priest/abilities/ShadowWordPain.kt
CyberQuackGames: Went with `DAMAGE`

src/commonMain/kotlin/character/classes/priest/buffs/InnerFocus.kt
CyberQuackGames: Went with `DAMAGE`

src/commonMain/kotlin/character/classes/priest/debuffs/ShadowWeavingDebuff.kt
CyberQuackGames: I updated Shadow Weaving to be a raid "buff" now to simplify this. I also synced the names by pulling in w/e name the priest buff version has for the raid buff so they'll be in sync.

src/commonMain/kotlin/character/classes/priest/abilities/ShadowWordPain.kt
CyberQuackGames: Using spell mechanics for all dots now.

src/commonMain/kotlin/character/races/abilities/DevouringPlague.kt
CyberQuackGames: Using spell mechanics for all dots now.

src/commonMain/kotlin/character/races/abilities/Starshards.kt
CyberQuackGames: Using spell mechanics for all dots now.

CyberQuackGames:
> Hey, Im a top 10 TBC Shadow Priest and "guide" writer for the priest discord, I'd love to work with you on making this as good as it could be. If you're interested add me on Discord @passion#6241

Hey there, I'm just about done implementing Shadow Priest here I believe (starting point anyway). I'm on the priest discord server as well and will post in the sim thread when I see this go live for us.

src/commonMain/kotlin/character/Proc.kt
CyberQuackGames: Removed the `SPELL_CAST` trigger again

wobbelbonk:
> 
> 
> > Hey, Im a top 10 TBC Shadow Priest and "guide" writer for the priest discord, I'd love to work with you on making this as good as it could be. If you're interested add me on Discord @passion#6241
> 
> Hey there, I'm just about done implementing Shadow Priest here I believe (starting point anyway). I'm on the priest discord server as well and will post in the sim thread when I see this go live for us.

Sim thread?
What sim thread?

src/commonMain/kotlin/character/classes/priest/debuffs/MindFlayDot.kt
CyberQuackGames: Added set bonuses

src/commonMain/kotlin/character/classes/priest/abilities/MindBlast.kt
CyberQuackGames: Added set bonuses

src/commonMain/kotlin/character/classes/priest/abilities/ShadowWordPain.kt
CyberQuackGames: Added set bonuses

src/commonMain/kotlin/character/races/abilities/Starshards.kt
CyberQuackGames: Converted to use HIT / CRIT

ui/src/presets/samples/priest_shadow_phase2.yml
CyberQuackGames: Looking at a few options here. I may just go with T5 as the sim is showing it is about equal with off pieces / engineering helm in shorter fights, but longer fights T5 wins out as well as what we can assume would be significantly better in spell cleave scenarios.

src/commonMain/kotlin/character/classes/priest/abilities/MindFlay.kt
CyberQuackGames: Updated for `DAMAGE` event

src/commonMain/kotlin/character/classes/priest/abilities/Shadowfiend.kt
CyberQuackGames: Added set bonuses now

src/commonMain/kotlin/character/classes/priest/abilities/MindFlay.kt
CyberQuackGames: Integrated with spell mechanics now.

src/commonMain/kotlin/character/classes/priest/abilities/VampiricTouch.kt
CyberQuackGames: Using spell mechanics for all dots now.

ui/src/presets/samples/priest_shadow_phase2.yml
CyberQuackGames: I put together variations of T5 and off pieces to include engineering goggles, but the latest push comes out on top in terms of a 3 minute fight. The longer the fight goes T5 is better. It's also likely better in spell cleave scenarios.

CyberQuackGames:
@marisa-ashkandi Ok if you want to review again here for last fixes and updates. I removed the spell damage taken stat and made Shadow Weaving a raid buff. In terms of dot snap shots, I've integrated the spell mechanics there. I added a method to `Spell.kt` for passing in a spell damage override as `baseDamageRoll` has a lot going on already on top of adding the `canCrit`, `canResist`, and `bonusDamageMultiplier` (needed for many set bonuses) to `attackRoll`. Let me know what your thoughts on that are, but it feels more consistent now in terms of events and damage calcs being the same on all dots.

Lastly I updated the phase 2 set to the highest 3 minute dps one I found through a bunch of comparisons with T5 and off pieces.

secretbis (approved):
Fabulous work, looks great!  I appreciate you taking the time to make all the changes!

PR #93, opened 2021-09-06
Fix Soulfrost enchant spell damage

CyberQuackGames:
https://tbc.wowhead.com/spell=46538/enchant-weapon-soulfrost


secretbis (approved):
Thanks for the fix!

PR #95, opened 2021-09-10
Add dungeon gems and a few meta updates

### LINKED ISSUE #28: Missing Gems
The following gems are missing from your sim. 

Deadly Fire Opal
Glinting Fire Opal
Shifting Tanzanite
(Only focused on Hunter related gems)

CyberQuackGames:
Adds [dungeon](https://tbc.wowhead.com/items/gems/quality:4?filter=128;4;0#0+14+1) and Bracing Earthstorm Diamond meta gems. Also fixed the Relentless Earthstorm Diamond not having a requirement check.

Resolves https://github.com/marisa-ashkandi/tbcsim/issues/28 as well.

Also I think this quote applies to this PR
"Well I got this far, I might as well keep going"... - Forrest Gump


src/commonMain/kotlin/data/gems/RelentlessEarthstormDiamond.kt
```diff
@@ -24,5 +25,13 @@ class RelentlessEarthstormDiamond : Gem(32409, "Relentless Earthstorm Diamond",
         }
     }
 
+    override fun metaActive(sockets: List<Socket>): Boolean {
```
secretbis: This is currently the default implementation for `metaActive` in the `Gem` superclass, shared with Destructive Skyfire Diamond.  

I'd be OK with either leaving this un-overridden in RED, or changing the `Gem` `metaActive` method to `abstract` and moving this implementation to both places, but the override here as written is redundant.

secretbis (changes requested):

src/commonMain/kotlin/data/gems/RelentlessEarthstormDiamond.kt
CyberQuackGames: Good to know, I had to dive a bit into that space and didn't immediately notice. I'll just remove this to keep it simple.

CyberQuackGames:
@marisa-ashkandi Fixed up the Relentless meta. Hopefully this PR helps a bit with a bunch of other classes as well.

secretbis (approved):

PR #96, opened 2021-09-10
Add Divine and Improved Divine Spirit to raid buffs

CyberQuackGames:
Pretty simple addition. In the event the user doesn't check the Divine Spirit buff they will still receive it from Improved Divine Spirit. The separate Divine Spirit buff still allows for adding the 50 spirit to test without Improved if desired.


secretbis (approved):
Looks great, thanks!

PR #97, opened 2021-09-10
Add BUFF_STACKS_GTE criteria check

CyberQuackGames:
Added the opposite of BUFF_STACKS_LTE. At the moment this has a use case with Shadow Weaving stacks for trinket / ability usage.

### DIFF
--- a/src/commonMain/kotlin/sim/rotation/criteria/BuffStacksGte.kt
+++ b/src/commonMain/kotlin/sim/rotation/criteria/BuffStacksGte.kt
@@ -1,0 +1,31 @@
+package sim.rotation.criteria
+
+import sim.SimParticipant
+import sim.config.RotationRuleCriterion
+import sim.rotation.Criterion
+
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/sim/rotation/criteria/BuffStacksGte.kt
```diff
@@ -0,0 +1,31 @@
...
+    } catch(e: Exception) {
+        logger.warn { "Field 'stacks' must be an integer for criterion $type" }
+        null
+    }
+
+    override fun satisfied(sp: SimParticipant): Boolean {
+        if(buff == null || stacks == null) return false
+
+        val state = sp.buffs[buff]?.state(sp)
+        return state == null || state.currentStacks >= stacks
```
secretbis: A missing buff (aka `null` `state` value) should return `false` for this criterion.

secretbis (changes requested):

CyberQuackGames:
@marisa-ashkandi Thanks for review and catch there, should be good now.

secretbis (approved):

PR #99, opened 2021-09-11
Add Malediction as raid debuff

CyberQuackGames:
Adds Malediction as a raid "debuff" (3% more spell damage to the four schools). The hidden buff only enhances Curse of Elements if present and is prioritized when checked over a Warlocks talent for it.


secretbis (approved):
LGTM!  Thank you!

PR #104, opened 2021-09-28
Add of shadow wrath enchanted items

CyberQuackGames:
Request from shadow priests for simming. I took the simple path here as I didn't see a way to import enchanted gear. Couple notes here though are the item ids are the base items so the UI won't show the enchant in the tool tip (Enchanted items have a underscore followed by the enchant number (e.g. `30684_9`) but items only have item ids as an int. Second, the EP values will be zero so the items, while searchable, won't be near the top as expected for shadow priests for earlier phases


secretbis:
That'll work!  Thanks!  I created a followup issue for EP for element-specific damage, in general - it should be pretty easy to make that per-spec.  https://github.com/marisa-ashkandi/tbcsim/issues/109

secretbis (approved):

PR #105, opened 2021-09-28
Fix Dark Iron Smoking Pipe trinket not applying buff

CyberQuackGames:
I saw the buff was in the itemprocs auto generated file, but not sure if the buffs themselves are required to be manually done. I saw the Icon of Silver Crescent and basically just verbatim copied since it's the same trinket with a different id / name.


secretbis:
I had to add the AP one as well, so I rolled this into: https://github.com/marisa-ashkandi/tbcsim/commit/adcbe2b5282fc1f39caa41d72c60b07f838486a0

PR #106, opened 2021-09-28
Fix ring enchants not stacking

CyberQuackGames:
Fixes ring enchants (buffs) not stacking due to naming collision.


secretbis (approved):
Nice find!  Thank you!

secretbis (approved):
Nice find!  Thank you!

PR #128, opened 2021-12-26
Improve Arcane Mage implementation

borbrunpagle:
Various fixes and improvements to arcane mage rotation and consumables.

Add Mage Armor to the mage ability list.

Improvements to the arcane mage EP calculation.

Restore the apparently missing ability to run EP for a chosen profile.

Instead of applying a random 2% hit reduction, which messes with various proc chances, etc. switch the delta on the hit EP test to use a negative hit modifier.

Add mage armor, which is used by arcane mages.

Add a spirit buff mutex and scroll of spirit.

Add Innervate.

Add missing elixirs used by arcane mages.

Frostbolt gets double benefit from Element Precision. (Who knows why?)

### DIFF
--- a/src/commonMain/kotlin/character/Spec.kt
+++ b/src/commonMain/kotlin/character/Spec.kt
@@ -12,7 +12,7 @@ abstract class Spec {
             Triple("strength", Stats(strength = 50), 50.0),
             Triple("agility", Stats(agility = 50), 50.0),
             Triple("meleeCritRating", Stats(meleeCritRating = 5.0 * Rating.critPerPct), 5.0 * Rating.critPerPct),
-            Triple("physicalHitRating", Stats(physicalHitRating = 2.0 * Rating.physicalHitPerPct), 2.0 * Rating.physicalHitPerPct),
+            Triple("physicalHitRating", Stats(physicalHitRating = -5.0 * Rating.physicalHitPerPct), -5.0 * Rating.physicalHitPerPct),
             Triple("physicalHasteRating", Stats(physicalHasteRating = 5.0 * Rating.hastePerPct), 5.0 * Rating.hastePerPct),
             Triple("expertiseRating", Stats(expertiseRating = 2.0 * Rating.expertisePerPct), 2.0 * Rating.expertisePerPct),
             Triple("armorPen", Stats(armorPen = 100), 100.0),
@@ -33,7 +33,7 @@ abstract class Spec {
         // AKA Enhancement Shaman
         val casterHybridDeltas = listOf(
             Triple("spellCritRating", Stats(spellCritRating = 5.0 * Rating.critPerPct), 5.0 * Rating.critPerPct),
-            Triple("spellHitRating", Stats(spellHitRating = 5.0 * Rating.spellHitPerPct), 5.0 * Rating.spellHitPerPct)
+            Triple("spellHitRating", Stats(spellHitRating = -5.0 * Rating.spellHitPerPct), -5.0 * Rating.spellHitPerPct)
         )
         val defaultCasterDeltas: List<SpecEpDelta> = listOf(
             Triple("intellect", Stats(intellect = 50), 50.0),
--- a/src/commonMain/kotlin/character/Spec.kt
+++ b/src/commonMain/kotlin/character/Spec.kt
@@ -14,7 +14,7 @@ abstract class Spec {
             Triple("meleeCritRating", Stats(meleeCritRating = 5.0 * Rating.critPerPct), 5.0 * Rating.critPerPct),
             Triple("physicalHitRating", Stats(physicalHitRating = -5.0 * Rating.physicalHitPerPct), -5.0 * Rating.physicalHitPerPct),
             Triple("physicalHasteRating", Stats(physicalHasteRating = 5.0 * Rating.hastePerPct), 5.0 * Rating.hastePerPct),
-            Triple("expertiseRating", Stats(expertiseRating = 2.0 * Rating.expertisePerPct), 2.0 * Rating.expertisePerPct),
+            Triple("expertiseRating", Stats(expertiseRating = -2.0 * Rating.expertisePerPct), -2.0 * Rating.expertisePerPct),
             Triple("armorPen", Stats(armorPen = 100), 100.0),
         )
 
@@ -23,7 +23,7 @@ abstract class Spec {
         val defaultRangedDeltas: List<SpecEpDelta> = listOf(
             Triple("agility", Stats(agility = 50), 50.0),
             Triple("rangedCritRating", Stats(rangedCritRating = 5.0 * Rating.critPerPct), 5.0 * Rating.critPerPct),
-            Triple("physicalHitRating", Stats(physicalHitRating = 2.0 * Rating.physicalHitPerPct), 2.0 * Rating.physicalHitPerPct),
+            Triple("physicalHitRating", Stats(physicalHitRating = -2.0 * Rating.physicalHitPerPct), -2.0 * Rating.physicalHitPerPct),
             Triple("physicalHasteRating", Stats(physicalHasteRating = 5.0 * Rating.hastePerPct), 5.0 * Rating.hastePerPct),
             Triple("armorPen", Stats(armorPen = 100), 100.0)
         )
@@ -37,7 +37,7 @@ abstract class Spec {
         )
         val defaultCasterDeltas: List<SpecEpDelta> = listOf(
             Triple("intellect", Stats(intellect = 50), 50.0),
-            Triple("spellHasteRating", Stats(spellHasteRating = 10.0 * Rating.hastePerPct), 10.0 * Rating.hastePerPct),
+            Triple("spellHasteRating", Stats(spellHasteRating = 5.0 * Rating.hastePerPct), 5.0 * Rating.hastePerPct),
         ) + casterHybridDeltas
     }
     abstract val name: String
--- a/src/commonMain/kotlin/character/classes/mage/abilities/Frostbolt.kt
+++ b/src/commonMain/kotlin/character/classes/mage/abilities/Frostbolt.kt
@@ -49,7 +49,7 @@ class Frostbolt : Ability() {
     val spellPowerCoeff = Spell.spellPowerCoeff(baseCastTimeMs)
     override fun cast(sp: SimParticipant) {
         val elementalPrecision: ElementalPrecision? = sp.character.klass.talentInstance(ElementalPrecision.name)
-        val emHit = elementalPrecision?.bonusFireFrostHitPct() ?: 0.0
+        val emHit = 2 * (elementalPrecision?.bonusFireFrostHitPct() ?: 0.0)
 
         val empFb: EmpoweredFrostbolt? = sp.character.klass.talentInstance(EmpoweredFrostbolt.name)
         val bonusFbCrit = empFb?.frostboltAddlCritPct() ?: 0.0
--- a/src/commonMain/kotlin/character/classes/mage/abilities/MageArmor.kt
+++ b/src/commonMain/kotlin/character/classes/mage/abilities/MageArmor.kt
@@ -1,5 +1,32 @@
-package character.classes.mage
+package character.classes.mage.abilities
 
-class MageArmor {
+import character.Ability
+import character.Buff
+import character.Stats
+import mechanics.General
+import sim.SimParticipant
 
+class MageArmor : Ability() {
+    companion object {
+        const val name = "Mage Armor"
+    }
+    override val id: Int = 22783
+    override val name: String = Companion.name
+    override val icon: String = "spell_magearmor.jpg"
+    override fun gcdMs(sp: SimParticipant): Int = sp.spellGcd().toInt()
+    override fun resourceCost(sp: SimParticipant): Double = 630.0
+
+    val buff = object : Buff() {
+        override val name: String = Companion.name
+        override val icon: String = "spell_magearmor.jpg"
+        override val durationMs: Int = 30 * 60 * 1000
+
+        override fun modifyStats(sp: SimParticipant): Stats {
+            return Stats(manaPer5Seconds = (General.mp5FromSpiritNotCasting(sp) * .3).toInt())
+        }
+    }
+
+    override fun cast(sp: SimParticipant) {
+        sp.addBuff(buff)
+    }
 }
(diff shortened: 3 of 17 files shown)


src/commonMain/kotlin/character/classes/mage/abilities/MageArmor.kt
```diff
@@ -0,0 +1,5 @@
+package character.classes.mage
+
+class MageArmor {
+
```
secretbis, line 8: Missing implementation?

src/commonMain/kotlin/character/classes/mage/specs/Arcane.kt
```diff
@@ -15,7 +18,7 @@ class Arcane : Spec() {
 
     override fun yellowSocketEp(deltas: Map<String, Double>): Double {
         // 5 spell haste rating / 6 spell damage
-        return ((deltas["spellHasteRating"] ?: 0.0) * 5.0) + 6.0
+        return ((deltas["intellect"] ?: 0.0) * 10.0)
```
secretbis, line 22: Nit: Can we update the comment above this line?

src/commonMain/kotlin/data/abilities/generic/AdeptsElixir.kt
```diff
@@ -0,0 +1,5 @@
+package data.abilities.generic
+
+class AdeptsElixir : Ability() {
+
```
secretbis, line 26: Missing implementation?

src/commonMain/kotlin/data/abilities/generic/ElixirOfDraenicWisdom.kt
```diff
@@ -0,0 +1,2 @@
+package data.abilities.generic 
+
```
secretbis, line 2: Missing implementation?

src/commonMain/kotlin/data/abilities/generic/Innervate.kt
```diff
@@ -0,0 +1,2 @@
+package data.abilities.generic 
+
```
secretbis, line 2: Missing implementation?

src/commonMain/kotlin/data/abilities/generic/ScrollOfSpiritV.kt
```diff
@@ -0,0 +1,2 @@
+package data.abilities.generic 
+
```
secretbis, line 2: Missing implementation?

src/commonMain/kotlin/data/abilities/raid/ImprovedDivineSpirit.kt
```diff
@@ -20,6 +21,7 @@ class ImprovedDivineSpirit : Ability() {
         override val name: String = Companion.name
         override val icon: String = "spell_holy_prayerofspirit.jpg"
         override val durationMs: Int = -1
+        override val mutex: List<Mutex> = listOf(Mutex.BUFF_SPIRIT)
```
secretbis: This is actually implemented with no overlap with regular Divine Spirit.  We can either remove this mutex, or keep the mutex and add the actual spirit bonus to the stats block on L28 below.

ui/src/presets/samples/mage_arcane_phase2.yml
```diff
@@ -112,14 +112,28 @@ gear:
...
+          resourceType: MANA
+        - type: FIGHT_TIME_REMAINING_GTE
+          seconds: 30
     - name: Blood Fury
     - name: Berserking
+    - name: Bloodlust
+      criteria:
+        - type: FIGHT_DURATION_GTE
+          seconds: 15
+    - name: Drums of Battle
```
secretbis, line 136: Drums are a configurable raid buff, and assumes it is cast by someone else - is the intent here to account for the drum cast time?  In any case, this wouldn't have any effect in a rotation, since it's not a registered `Ability` for the mage class.

ui/src/presets/samples/mage_arcane_phase2.yml
```diff
@@ -112,14 +112,28 @@ gear:
...
+    - name: Evocation
+      criteria:
+        - type: RESOURCE_PCT_LTE
+          pct: 30
+          resourceType: MANA
+        - type: FIGHT_TIME_REMAINING_GTE
+          seconds: 30
     - name: Blood Fury
     - name: Berserking
+    - name: Bloodlust
```
secretbis, line 132: Bloodlust is also a configurable raid buff, and is not an ability Mages can cast.  As such, it should be removed from the rotation, as it won't do anything here.

ui/src/presets/samples/mage_arcane_phase2.yml
```diff
@@ -112,14 +112,28 @@ gear:
 rotation:
   autoAttack: false
   precombat:
-    - name: Flask of Blinding Light
+    - name: Elixir of Draenic Wisdom
+    - name: Adept's Elixir
     - name: Crunchy Serpent
     - name: Arcane Intellect
-    - name: Molten Armor
+    - name: Mage Armor
```
secretbis, line 119: Would it be useful to conditionally cast Mage/Molten Armor based on fight duration?

src/jvmMain/kotlin/Main.kt
```diff
@@ -265,7 +258,11 @@ class TBCSim : CliktCommand() {
         val specFilter = specFilterStr?.split(",")
         val categoryFilter = categoryFilterStr?.split(",")
 
-        if (calcEP) {
+        if (calcEP && configFile?.exists() == true) {
```
secretbis, line 261: Sorry about this, looks like I broke that option a long time ago and never fixed the docs...  The EP rankings on the site mostly replaced the intent of that feature.  

Would the most useful output for a single-character sim be just a table of results to the console?  Or, would something else be better?

src/commonMain/kotlin/character/classes/mage/abilities/Frostbolt.kt
```diff
@@ -49,7 +49,7 @@ class Frostbolt : Ability() {
     val spellPowerCoeff = Spell.spellPowerCoeff(baseCastTimeMs)
     override fun cast(sp: SimParticipant) {
         val elementalPrecision: ElementalPrecision? = sp.character.klass.talentInstance(ElementalPrecision.name)
-        val emHit = elementalPrecision?.bonusFireFrostHitPct() ?: 0.0
+        val emHit = 2 * (elementalPrecision?.bonusFireFrostHitPct() ?: 0.0)
```
secretbis, line 52: Blizzard works in mysterious ways...

src/commonMain/kotlin/character/Spec.kt
```diff
@@ -12,7 +12,7 @@ abstract class Spec {
             Triple("strength", Stats(strength = 50), 50.0),
             Triple("agility", Stats(agility = 50), 50.0),
             Triple("meleeCritRating", Stats(meleeCritRating = 5.0 * Rating.critPerPct), 5.0 * Rating.critPerPct),
-            Triple("physicalHitRating", Stats(physicalHitRating = 2.0 * Rating.physicalHitPerPct), 2.0 * Rating.physicalHitPerPct),
+            Triple("physicalHitRating", Stats(physicalHitRating = -5.0 * Rating.physicalHitPerPct), -5.0 * Rating.physicalHitPerPct),
```
secretbis, line 15: This is actually a lot cleaner!  However, we would need to update all entries for `physicalHitRating`, `expertiseRating`, and `spellHitRating` to use the negative modifiers.  Those were all the same idea.

secretbis (changes requested):
Thanks for the Mage expertise!  I don't play one myself, so this is very helpful!

I noticed a few missing implementations - are there some missing commits for this PR?  Let me know if there is anything I can help with!

ui/src/presets/samples/mage_arcane_phase2.yml
borbrunpagle: It wasn't clear to me how to configure the Bloodlust buff. For example, how do you specify when it triggers?

ui/src/presets/samples/mage_arcane_phase2.yml
borbrunpagle: Not really. You basically use Mage Armor all the time.

src/jvmMain/kotlin/Main.kt
borbrunpagle: I found creating the list of my personal EPs to be the most useful thing. Arcane mages have a few different points where haste, in particular, changes very dramatically in value. Understanding where I am relative to those points of inflection is what I wanted to get, and the current output with my local changes does that.

src/commonMain/kotlin/character/Spec.kt
borbrunpagle: Done

src/commonMain/kotlin/character/classes/mage/abilities/MageArmor.kt
borbrunpagle: Fixed

src/commonMain/kotlin/character/classes/mage/specs/Arcane.kt
borbrunpagle: Fixed

src/commonMain/kotlin/data/abilities/generic/AdeptsElixir.kt
borbrunpagle: Fixed

src/commonMain/kotlin/data/abilities/generic/ElixirOfDraenicWisdom.kt
borbrunpagle: fixed

src/commonMain/kotlin/data/abilities/generic/Innervate.kt
borbrunpagle: Added implementation. Don't know how to figure out how much baseline spirit mp5 the character has though. I modeled it conservatively assuming the character has 60% base regen, which is what arcane mages have and is also the highest available in tbc.

src/commonMain/kotlin/data/abilities/generic/ScrollOfSpiritV.kt
borbrunpagle: fixed

src/commonMain/kotlin/data/abilities/raid/ImprovedDivineSpirit.kt
borbrunpagle: I added the mutex to allow spirit scroll vs divine spirit. I can take it out of here, but leave it for the those exclusive buffs.

src/commonMain/kotlin/data/abilities/generic/FlameCap.kt
```diff
@@ -0,0 +1,33 @@
...
+    override val sharedCooldown: SharedCooldown = SharedCooldown.RUNE_OR_MANA_GEM
+    override fun cooldownMs(sp: SimParticipant): Int = 180000
+
+    val buff = object : Buff() {
+        override val name: String = Companion.name
+        override val icon: String = Companion.icon
+        override val durationMs: Int = 60000
+
+        override fun modifyStats(sp: SimParticipant): Stats {
+            return Stats(fireDamage = 80)
```
secretbis, line 26: This is missing the melee proc implementation, but I'm happy to add that later - just noting so I don't forget.

src/commonMain/kotlin/data/abilities/generic/Innervate.kt
secretbis: Sounds good!  Not hard to tweak it later if it is too strong or too weak.  

It's possible that the allowed spirit-based regen modifier needs to be a Stat, so various abilities can collectively modify it more easily.

ui/src/presets/samples/mage_arcane_phase2.yml
secretbis: The raid version of Bloodlust is cast at the start of the fight, and lines up with generally-immediate cooldown usage by presets.

src/commonMain/kotlin/data/abilities/generic/ScrollOfSpiritV.kt
```diff
@@ -0,0 +1,30 @@
...
+    override val id: Int = 27501
+    override val name: String = Companion.name
+    override val icon: String = "inv_scroll_01.jpg"
+    override fun gcdMs(sp: SimParticipant): Int = 0
+
+    val buff = object : Buff() {
+        override val name: String = "Scroll of Spirit V"
+        override val icon: String = "inv_scroll_01.jpg"
+        override val durationMs: Int = 30 * 60 * 1000
+        override val mutex: List<Mutex> = listOf(Mutex.BUFF_SPIRIT)
```
secretbis, line 20: It would be useful to add an override for `mutexPriority` here, since that will resolve based on the potency of the buff.  The default priority is whichever is newer, which isn't correct for most stat-based buffs.

The value there would be `mapOf(Mutex.BUFF_SPIRIT to 30)`, and the corresponding `50` weight in Divine Spirit.

src/commonMain/kotlin/data/abilities/raid/DivineSpirit.kt
```diff
@@ -20,6 +21,7 @@ class DivineSpirit : Ability() {
         override val name: String = Companion.name
         override val icon: String = "spell_holy_prayerofspirit.jpg"
         override val durationMs: Int = -1
+        override val mutex: List<Mutex> = listOf(Mutex.BUFF_SPIRIT)
```
secretbis, line 24: As with the spirit scroll implementation, this would benefit from a `mutexPriority` override to accurately express the potency.

src/commonMain/kotlin/data/abilities/raid/ImprovedDivineSpirit.kt
```diff
@@ -2,6 +2,7 @@ package data.abilities.raid
 
 import character.Ability
 import character.Buff
+import character.Mutex
```
secretbis, line 5: Nit: Leftover import

src/jvmMain/kotlin/Main.kt
secretbis: Sounds good!  Whenever you're ready, I'd love to see the implementation!

secretbis (commented):
This is looking great!  I added a couple suggestions about spirit buff priority, and barring a couple comments in the rotation and the personal EP output, should be basically good to go soon!

secretbis:
Squashed in separately and completed

